Issue #23, opened 2022-11-03
Property menu/page does not work


koeaw:
**Describe the bug**
The `Relations > Property` menu/Property list view does not currently work. It is simply a duplicate of the Triple/Temptriple pages.

**To Reproduce**
1. Go to https://YOUR_APIS_HOST/apis/relations/property/list/
2. Compare with https://YOUR_APIS_HOST/apis/relations/triple/list/

**Expected behavior**
A table containing an Ontologies app's **properties** (subject-predicate-object relationships) should be shown.

koeaw:
It does not work 100% like I would like it to, but what I've got so far is very much usable and a lot more helpful than _no_ Property menu:

![properties_table_implemented](https://user-images.githubusercontent.com/113020169/199724093-2baacffa-bb9e-40a4-9e52-d4e9f8345b54.png)

Will send a PR once I've cleaned up my code.

PR #25, opened 2022-12-16
Best-effort fix of non-functioning Properties menu

koeaw:
**Describe your changes**
PR turns `Relations > Properties`  menu into a page that visualises the types of relationships possible/allowed between entity classes.

Some code comments suggest there are plans to rework how Triples are handled, so as not to interfere with that, no attempt at consolidating code to handle Triples and Properties was made; instead, the behaviour for Triples was mimicked for Properties. 

Summary:
- create new `class PropertyFilter`
- create new `class PropertyTable` with columns Subject, Predicate, Object, Reverse predicate
- add attributes to tables (requires fixing of templates to come into effect)
- explicitly specify fields/table columns to initially sort by in views
- add "id" column to Triple view (which is what Triple tables are default sorted by anyway, but isn't a sort order users can currently go back to once they sort by a different column)
- reformat files with black, clean up imports

**Screenshot of resulting table**
properties_table

**Caveats**
Relationships with more than one subject or more than one object result in duplicate rows even though each row already lists all subjects/objects. Attempts on my part to split them out into separate rows weren't successful.

properties_table_duplicate_rows

Counting of rows also is buggy/seems affected negatively by the same underlying issue: in the app I work on, sorting by Subject yields a different number (18) than sorting by Object (21). Some rows also appear to be missing when sorting in one direction, while others are missing when sorting in the other.

This solution is probably still better than not having a functioning Property page at all, so I'd suggest merging the PR and making sure everyone's aware of these bugs. (Hence: "best-effort fix".)

**Related issues and PRs**
Resolves issue:
- #23 


koeaw:
Actually, I forgot about something that I'd stashed away, so will set this PR to Draft for another day to revise it.

koeaw:
Ok, updated, incl. with Docstrings & code comments that had been stashed away, now ready to merge.

To expand on earlier notes:  
I haven't been able to figure out how to get at the **verbose names** of entities _and_ sort them alphabetically. My research suggests there's a way to retrieve more info about a model used in a ManyToMany field (`subj_class`, `obj_class` in `Property`) by using a `through` table, but that is not compatibel with `limit_choices_to`. + The  `Property` class is also used to build triples, and I didn't know how messing with the fields in the class would/could potentially affect those.

I'm also not familiar enough with Django Filters to know if those would make it possible to get at the verbose names (hence also why `PropertyFilter` is a 1:1 copy of `TripleFilter`). Same goes for the `filter` attribute of the django_tables2 `ManyToManyColumn`s, obviously.

This revision now results in the correct number of rows being output for the properties themselves, in any case.

steffres:
Looks good. The triples and properties table are certainly an improvement. Thank you!

PR #29, opened 2023-01-17
Add a black linter action

b1rger:
This action looks at pull requests and checks the changed
Python files for code style conformity.


b1rger:
implemented as part of #42, therefore closing

PR #31, opened 2023-02-01
Allow overriding the APIS_ENTITIES setting

b1rger:
The APIS_ENTITIES are created using the values that are stored in
the `entity_settings` attribute of entities. In some cases it might
make sense to be able to define those in the global settings file.
This commit changes the code so it checks for the existense of the
APIS_ENTITIES setting and only uses the `entity_settings` attribute
if the global settings is empty.

### DIFF
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -183,7 +183,7 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
             "columns"
         )  # populates "Select additional columns" dropdown
         try:
-            default_cols = self.entity_settings["table_fields"]
+            default_cols = settings.APIS_ENTITIES.get(class_name)["table_fields"]
         except KeyError as e:
             default_cols = []  # gets set to "name" in get_entities_table when empty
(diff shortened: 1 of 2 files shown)


koeaw:
Relatedly, I think it would make sense if `entity_settings` were set up differently in general, see: https://github.com/acdh-oeaw/apis-ontologies/issues/9#issuecomment-1334010750 cc @sennierer

apis_core/apis_entities/views.py
```diff
@@ -183,7 +183,7 @@ def get_table(self, **kwargs):
             "columns"
         )  # populates "Select additional columns" dropdown
         try:
-            default_cols = self.entity_settings["table_fields"]
+            default_cols = settings.APIS_ENTITIES.get(class_name)["table_fields"]
```
sennierer, line 186: here the settings were again replaced for the ones from models.py. given that in models.py we use either the one or the other to set the settings, it should read here from the settings rather than the models.py

sennierer (approved):
works in my setup, with the additional commit

b1rger:
squashed, rebased and merged

Issue #33, opened 2023-02-13
Get rid of hard coded apis_ontology dependency


b1rger:
`apis_ontology` is a separate module, but there are various places in `apis-core-rdf` that import that module unconditionally (i.e. https://github.com/acdh-oeaw/apis-core-rdf/blob/3f2a227994d1581df42794077f60ef2974d34f7a/apis_core/apis_entities/models.py#L257). This makes it impossible to run tests in this repository, because there is no `apis_ontology` in this repository. It would be great if we could somehow make these imports check for the the existence of `apis_ontology`, so that all the tests that don't need this module can run without it.

PR #34, opened 2023-02-13
Work around hard coded dependency on apis_ontology

b1rger:
This commit introduces some try-except blocks around hardcoded imports
of `apis_ontology`. `apis_ontology` is not part of this project and
there is no dependency defined anywhere. With the exception handling
it becomes possible to run tests in this project.

Closes: https://github.com/acdh-oeaw/apis-core-rdf/issues/33

### DIFF
--- a/apis_core/api_routers.py
+++ b/apis_core/api_routers.py
@@ -4,7 +4,10 @@ import importlib
 import inspect
 from ast import literal_eval
 from django.conf import settings
-from apis_ontology.models import *
+try:
+    from apis_ontology.models import *
+except ImportError:
+    pass
 from apis_core.apis_metainfo.models import *
 # from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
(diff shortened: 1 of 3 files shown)


steffres:
What use-case does this PR cover? 

The data model of a project is defined in `apis_ontology.models`. Without it there wouldn't be any entities and so I see no immediate point in running an apis instance without them. 

So why bother about the case where the module is not present?

steffres:
Ah, that's why: https://github.com/acdh-oeaw/apis-core-rdf/issues/33

Sure then.

b1rger:
This is a draft. You can ignore it.

steffres:
Looks fine. But I didn't test it because this PR serves the purpose of testing as outlined in the related issue. And since these tests do not yet exist, I didn't bother running a module on its own. 

The changes are minimalistic anyway and they won't break anything in production, so feel free to go ahead with merging.

PR #45, opened 2023-02-21
apis_entities: use defaults if entity not in settings.APIS_ENTITIES

b1rger:
If the entity is not defined in the settings.APIS_ENTITIES dict, we use a
default of an empty dict instead of throwing an exeption.

### DIFF
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -182,10 +182,10 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
         selected_cols = self.request.GET.getlist(
             "columns"
         )  # populates "Select additional columns" dropdown
-        try:
-            default_cols = settings.APIS_ENTITIES.get(class_name)["table_fields"]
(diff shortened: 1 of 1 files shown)


koeaw (commented):
I like this replacement, but would stick to `entity_settings` instead of `class_settings` based on usage elsewhere (solely for consistency; helpfulness of var name is a diff. discussion to be had). + I think it should be double quotation marks according to Black, though of little relevance considering the upcoming PR for reformatting the entire codebase.

koeaw:
General note:

What the variable name has made me aware of is that previous refactorings of mine (for `entity_settings` saved in models) don't make as much sense anymore if we want to switch (back) to using `APIS_ENTITIES`, and that `APIS_ENTITIES` should probably be checked further up already like I did with `entity_settings`. (And maybe that renders the argument about the variable name irrelevant after all.)

It might make sense to go over this in person because I refactored for different goals in different places all at once, on top of not understanding the underlying structure 100% (and trying to untangle that even gives me a headache now).

PR #49, opened 2023-02-28
apis_entities: refactor get_all_entity_classes

b1rger:
This commit refactors the `get_all_entity_classes()` method. Instead of loading a
hardcoded module we simply get a list of all models and check wether they are
children of TempEntityClass (or in the future of another class that is the root
of all entities). This way we can also loose the inspect import.

I tested the changes in an test instance. Before merging this PR I would like
to clarify the reason of or remove the `ent_class` comparison.

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -1,4 +1,3 @@
-import inspect
 import re
 import sys
 import unicodedata
@@ -233,8 +232,8 @@ class AbstractEntity(RootObject):
     # Methods dealing with all entities
     ###########################################################################
 
-    _all_entity_classes = None
-    _all_entity_names = None
+    _all_entity_classes = []
+    _all_entity_names = []
 
     @classmethod
     def get_all_entity_classes(cls):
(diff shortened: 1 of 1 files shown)


steffres:
We can merge this PR for now. Didn't run it, but from looking at the code I trust that this works.

But just as a heads-up: I already see these following problems will emerge with the upcoming restructuring despite it:

* TempEntityClass will be removed from apis core logic and only assumed to exist in legacy ontologies. But for the future we want to move away from it entirely, so better not assume it to "represent an entity".
* RootObject will be the major "thing" for everything that exists in an ontology. However RootObject will also be used for Properties and Vocabularies / Concepts. So RootObject can't be assumed to stand for an entity, as a potential replacement. 
* We will definitely need a simple elegant way to differentiate between "Entity", "Property", "Vocabulary/Concept" since they will all inherit from RootObject. Maybe boolean flags attached to dedicated Abstract Classes for each? But that's a separate issue.

So for now, this PR will work, but it will need surely again little work in the near future.

b1rger:
I merged this now, but the question still stands: does anyone now what that `ent_class` refers to?

steffres:
I introduced the `ent_class` but I don't remember myself. Safe to throw away by now though.

steffres:
I pulled this to my local instance and now `./manage.py makemigrations` fails with `django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.`

@b1rger can you make migrations?

koeaw:
@b1rger I don't know either, I think I just put it in the Docstring for completeness' sake.

b1rger:
> I pulled this to my local instance and now `./manage.py makemigrations` fails with `django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.`
> 
> @b1rger can you make migrations?

Ah, yeah, that makes sense - the app registry is only loaded later. Given that the class will be remove anyway, I've created a PR that reverts the commit: https://github.com/acdh-oeaw/apis-core-rdf/pull/58 I'll look into how to solve that when we move to the new class hierarchy

steffres:
sensible, thank you.

koeaw:
`ent_class`  might be a leftover from vanilla apis-core? https://github.com/acdh-oeaw/apis-core/blob/main/apis_core/context_processors/custom_context_processors.py#L12

steffres:
> `ent_class` might be a leftover from vanilla apis-core? https://github.com/acdh-oeaw/apis-core/blob/main/apis_core/context_processors/custom_context_processors.py#L12

might be. Whatever it is, we can probably safely delete it.

PR #61, opened 2023-03-06
Introduce CI tests

b1rger:
- Clean up settings file for CI tests
- apis_entities: drop old unused tests
- Introduce test workflow


koeaw:
I can't comment on these changes/suggestions.

steffres:
I haven't worked on our CI/CD on github at all as we had migrated over in autumn from gitlab.com and since it was clear for me to phase out of the APIS project already back then I didn't bother looking into this. 

So no idea what that yml does. 

The removal of outdated classes can't be wrong however.

Perhaps @sennierer would be a better reviewer or could you suggest someone else?

b1rger:
> The removal of outdated classes can't be wrong however.

Thanks, my need for review regarding this PR was mostly about the removal of the old tests. I'll merge it once we have a `dev` where the tests don't fail

PR #64, opened 2023-03-08
Filters entities shown on relation editing panel

richardhadden:
Filters list of entities shown in relation panels to only those that can be subject/object of relations to which entity viewed is object/subject

**Describe your changes**
Edit page in right hand panel previously showed all entities, not just those to which a relation can actually be created.

Replaced `AbstractEntity.get_all_entity_classes()` (which gets all entities) with new function `get_related_classes`, which: 

- finds all Properties for which the entity is a subject and gets all the Properties' allowed objects
- finds all Properties for which the entity is an object and gets all the Properties' allowed subjects
- and is wrapped with `functools.lru_cache` to prevent the db queries on every request (in lieu of a better way, such as constructing a lookup when the Property is actually created)

### DIFF
--- a/apis_core/apis_entities/views2.py
+++ b/apis_core/apis_entities/views2.py
@@ -25,14 +25,41 @@ from .views import get_highlighted_texts
 from .views import set_session_variables
 from ..apis_vocabularies.models import TextType
 
+from apis_core.apis_relations.models import Property
+import itertools
+import functools
+
 if 'apis_highlighter' in settings.INSTALLED_APPS:
     from apis_highlighter.forms import SelectAnnotatorAgreement
+
+
+@functools.lru_cache
+def get_related_classes(entity_name: str):
+    
+    # Find all the properties where the entity is either subject or object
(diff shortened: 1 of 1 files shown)


koeaw:
Pst, @richardhadden, issues need to be manually added to the GitHub project for now to show up there (sidebar on the right). (The action for that isn't finished yet.)

steffres:
Sorry Richard, I had already implemented this just in the recent refactoring which was merged yesterday: 

Here iterating over Entity and Reification classes: https://github.com/acdh-oeaw/apis-core-rdf/blob/e8bb50a17eb108f9427cd77c5fed156b570da8df/apis_core/apis_entities/views2.py#L50
and here filtering for related classes https://github.com/acdh-oeaw/apis-core-rdf/blob/e8bb50a17eb108f9427cd77c5fed156b570da8df/apis_core/apis_entities/views2.py#L54

And I prefer to keep my implementation as it's less code than yours. Unless you or others see another reason for yours?

Thanks though!

PR #74, opened 2023-03-10
Fix double import and reintroduce lost change

b1rger:
This fixes a double import that was introduced in e90f0a8 and reintroduces a
change from 90a1ad5 that got lost.


steffres:
damnit. Yeah, that's better. Thx.

steffres:
free to merge.

PR #76, opened 2023-03-10
Replace self_contenttype with property

b1rger:
Given that there is a cache now, we can use that to get the contenttype of the
class from memory instead of storing this information in the database every
instance of a model.
That way we don't use the database that often.

@steffres, what do you think? If that approach makes sense, I would update the
rest of the code that uses the `self_contenttype` field.

### DIFF
--- a/apis_core/apis_metainfo/models.py
+++ b/apis_core/apis_metainfo/models.py
@@ -342,17 +342,8 @@ class RootObject(models.Model):
     # self_contenttype: a foreign key to the respective contenttype comes in handy when querying for
     # triples where the subject's or object's contenttype must be respected (e.g. get all triples
     # where the subject is a Person)
-    self_contenttype = models.ForeignKey(
-        ContentType, on_delete=models.deletion.CASCADE, null=True, blank=True
-    )
     objects = models.Manager()
     objects_inheritance = InheritanceManager()
-
-    def save(self, *args, **kwargs):
-        if self.self_contenttype is None:
-            self.self_contenttype = caching.get_contenttype_of_class(self.__class__)
-
-        super().save(*args, **kwargs)
 
     def __str__(self):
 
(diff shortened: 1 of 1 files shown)


steffres:
Sadly not feasible. I had considered that too before. But the critical downside of this would be that you couldn't use it in Django's ORM anymore, like so: ` Triple.objects.filter(subj__self_contenttype=some_ct)`, same for the REST API where we have even less control within the python runtime.

Persisting the contenttype into each row of an entity table is wasteful sadly, and checking for it requires a DB hit, but the advantage of full integration into the ORM and overall consistency of its usage outweighs the drawbacks imo.

Unless there is some other approach where we could have the best of both worlds? I can't think of one though, other than perhaps overriding the general Django ORM manager for this. But that would probably break more than it could help.

b1rger:
> Sadly not feasible. I had considered that too before. But the critical downside of this would be that you couldn't use it in Django's ORM anymore, like so: ` Triple.objects.filter(subj__self_contenttype=some_ct)`, same for the REST API where we have even less control within the python runtime.
> 
> Persisting the contenttype into each row of an entity table is wasteful sadly, and checking for it requires a DB hit, but the advantage of full integration into the ORM and overall consistency of its usage outweighs the drawbacks imo.
> 
> Unless there is some other approach where we could have the best of both worlds? I can't think of one though, other than perhaps overriding the general Django ORM manager for this. But that would probably break more than it could help.

Well, for the ORM you can do for example `TempEntityClass.objects.filter(person=True)`. I looked a bit at the REST API and `api_router.py` and there I was able to filter against `OneToOneRel` type fields using `isnull`, so filtering persons was also possible.

Not sure how the late changes in inheritance relate to this though.

steffres:
Can I close this PR Bircher, or do you want to keep it open should you come up with an alternative approach?

Issue #77, opened 2023-03-10
Replace `title()` uses on entity classes


SteffRhes:
Throughout the codebase there is legacy use of `title()` on entity class names. All of them need to replaced. Also the usage of it must be avoided in the future. 

It was be used in two contexts where for both of them are problematic:

* If `title()` is used to identify a class (e.g. as key in a dictionary), then it's highly unstable as it screws up camel-cased class names (e.g. `"BookReview".title()` becomes `Bookreview`) and we don't want to break with Python's class naming convention either by making them lowercase.

* If it's used for "nice" displaying of class names in UI, then `verbose_name` should be used instead. 

I can't think of any other use-case than those two. And since we should not use it in those two, we should not use `title()` on entity class names at all anymore.

Pycharm tells me of 44 occurrences of `title()`, excluding commented code.

koeaw:
Almost. As said elsewhere, there's a point to `.title()` when it's used on verbose names of classes, since those are assumed to be lower case.

steffres:
`title()` is used by Django on verbose names, yes. But we are talking here about replacing our own internal usage of it, which should be dropped.

koeaw:
No, I was specifically talking about our own use, used by e.g. me [here](https://github.com/acdh-oeaw/apis-core-rdf/blob/dev/apis_core/apis_entities/views.py#L235), which is only conveniently clear because I chose an unambiguous variable name.

My comment was targeted at whoever ends up picking up this issue so they don't take this:

> All of them need to replaced. Also the usage of it must be avoided in the future.

at face value because it could lead to them accidentally removing those valid/useful cases of `.title()`-ing on verbose names as well.

steffres:
It's better to set `verbose_name` manually for classes like so:
```
class E40_Legal_Body(E1_Crm_Entity):
    class Meta:
        verbose_name = "Legal Body"
```
This way we can define within the ontology one precise label to be displayed for humans, and we can separate it from the technical model class name.

Hence, no need for `title()`. Just use `model._meta.verbose_name` alone.

koeaw:
_This is already happening._

> As said elsewhere, there's a point to .title() when it's used on verbose names of classes, since those are assumed to be lower case.

b1rger:
After #1015 there will only be one occurrence of `title()` left, namely:
[...]

PR #103, opened 2023-03-21
apis_entities: fix merging two entities

b1rger:
The RDFParser needs the correct app_label_entities set, to find the
ContentType. We could also fix the RDFParser directly, but at the
current state this could have unknown consequences.
In the merge method of the TempEntityClass it seems that the lookup
of the ContentType also works withtout setting the app_label.

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -283,9 +283,7 @@ class TempEntityClass(AbstractEntity):
         from apis_core.apis_metainfo.models import Uri
 
         e_a = type(self).__name__
-        self_model_class = ContentType.objects.get(
-            app_label="apis_entities", model__iexact=e_a
-        ).model_class()
+        self_model_class = ContentType.objects.get(model__iexact=e_a).model_class()
         if isinstance(entities, int):
             entities = self_model_class.objects.get(pk=entities)
         if not isinstance(entities, list) and not isinstance(entities, QuerySet):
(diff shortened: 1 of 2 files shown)


apis_core/apis_entities/models.py
```diff
@@ -778,9 +778,7 @@ def merge_with(self, entities):
         from apis_core.apis_metainfo.models import Uri
 
         e_a = type(self).__name__
-        self_model_class = ContentType.objects.get(
-            app_label="apis_entities", model__iexact=e_a
-        ).model_class()
+        self_model_class = ContentType.objects.get(model__iexact=e_a).model_class()
```
steffres, line 781: Here is a DB call on `ContentType`, and then on the received instance `model_class()`. 

This could be easily replaced by using our caching module like so `caching.get_entity_class_of_name(e_a)`

b1rger: Thanks for the comment. We will keep that in mind, a separate PR with a global introduction of the caching module wherever possible would probably be more appropriate. Another approach is to hide the caching implementation details by simply providing a `get_class` method in the entity.

steffres: I agree that a global replacement PR would be better. I made an issue: https://github.com/acdh-oeaw/apis-core-rdf/issues/108

And while we could attach a method to a model class, I'm sceptical. What's the benefit of it in contrast to using the caching module? I see it rather as redundancy and minor inconsistency. 

Calling such a function through a class method does not hide more implementation than calling function itself already imo.

koeaw:
Skimming through the log I couldn't help but notice the `entity.title()` that's part of this change, which suspiciously looks like it could be a leftover case of referencing a class by title-casing its slug rather than using its actual model name (or its verbose name for display purposes), so thought I'd flag it up.

[Issue summarising the problem](https://github.com/acdh-oeaw/apis-core-rdf/issues/77) which I've commented on in older PRs that fixed this issue in other places.

PR #107, opened 2023-03-30
Improve how "source" data is displayed on admin frontend

koeaw:
**Describe your changes**
* Add  **Source**  to an entity's detail view template.
* Always use a source's name (`orig_filename` field) if it is set, don't default to only ID just because the field for author info is empty.

**Additional context**
This class needs documentation added to it to explain its purpose and to make its fields clearer. To me it seems the fields are in need of a general overhaul. -> They weren't super useful for our purposes, and some changes in this PR work around that problem.

**Checklist**
- [x] I have added new unit tests or updated existing ones to demonstrate my changes works

### DIFF
--- a/apis_core/apis_metainfo/models.py
+++ b/apis_core/apis_metainfo/models.py
@@ -88,6 +88,8 @@ class Source(models.Model):
     def __str__(self):
         if self.author != "" and self.orig_filename != "":
             return "{}, stored by {}".format(self.orig_filename, self.author)
+        elif self.orig_filename != "":
+            return "{}".format(self.orig_filename)
         else:
             return "(ID: {})".format(self.id)
 
--- a/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
+++ b/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
@@ -185,6 +185,14 @@
                 <table class="table table-bordered table-hover">
                   <tr>
                     <th>
+                      Source
+                    </th>
+                    <td>
+                      {{ object.source }}
+                    </td>
+                  </tr>
+                  <tr>
+                    <th>
                       Collection(s)
                     </th>
                     <td>
--- a/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
+++ b/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
@@ -5,329 +5,329 @@
 {% block scriptHeader %}
 {{ block.super }}
 {% if tei %}
-    <link rel="stylesheet" href="{{ ceteicean_css }}" media="screen" charset="utf-8">
-    <script src="{{ ceteicean_js }}"></script>
+  <link rel="stylesheet" href="{{ ceteicean_css }}" media="screen" charset="utf-8">
+  <script src="{{ ceteicean_js }}"></script>
 {% endif %}
 {% if iiif %}
-    <script src="{{ openseadragon_js }}"></script>
+  <script src="{{ openseadragon_js }}"></script>
 {% endif %}
 {% endblock scriptHeader %}
 {% block content %}
 <div class="container-fluid">
-    <div class="card">
-        <div class="card-header">
-            <div class="row">
-                <div class="col-md-2">
-                    {% if object.get_prev_url %}
-                        <h2>
-                            <a href="{{ object.get_prev_url }}" style="float:right">
-                                <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-chevron-left">
-                                    <polyline points="15 18 9 12 15 6" />
-                                </svg>
-                            </a>
-                        </h2>
-                    {% endif %}
-                </div>
-                <div class="col-md-8">
-                    <h2 style="text-align: center;">
-                        {% if object.get_listview_url %}
-                            <a href="{{ object.get_listview_url }}">
-                                <small>{{ entity_type }}s</small>
-                            </a>
-                        {% else %}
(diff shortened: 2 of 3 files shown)


steffres:
`Source` is a legacy class from vanilla and should be removed. If a `Source` class is needed in a project it better be defined in an ontology.

Sorry to say, but that will make this work obselete. 

Did you coordinate this with Matthias or Matej beforehand?

koeaw:
@steffres In which document or ticket does it say this class is going to be removed?

Two months ago, I specifically asked about the purpose of `Collection` vs. `Source` on MM and _we rewrote our import scripts_ from using the former to using the latter based on the feedback I received there.

I'm aware the field that's currently part of `TempEntityClass` is not going to continue to exist as such, but I wasn't aware the class was going to be dropped altogether.

koeaw:
> I didn't look into the code of this PR, but how hard do you estimate will it be to remove later again? Because if it's something that you need now and since the removal of these classes probably still take a while as it's low priority it might make sense to use your PR for this limited time window nevertheless.

The relevant changes are minor:

* expanded if-statement in `save` method: https://github.com/acdh-oeaw/apis-core-rdf/pull/107/commits/4afffbdd306de051add2510d39d232bcd2f56d03
* added `Source` to a template: https://github.com/acdh-oeaw/apis-core-rdf/pull/107/commits/d7cdc32ac445b9554c8866918a3ba7b89c28f921

The current implementation is unfortunately useless to us because it only shows IDs for sources.

I've already deployed the changes for our project like I always do, but getting them merged into the shared branch _very much reduces overhead_ (keeping my branches in sync with upstream is lots of extra work).

steffres:
I see. Sure, looks good from my side.

apis_core/apis_metainfo/models.py
```diff
@@ -88,6 +88,8 @@ class Source(models.Model):
     def __str__(self):
         if self.author != "" and self.orig_filename != "":
             return "{}, stored by {}".format(self.orig_filename, self.author)
+        elif self.orig_filename != "":
+            return "{}".format(self.orig_filename, self.author)
```
b1rger, line 92: I think f-strings are more readable: `f"{self.orig_filename}"` - and it doesn't really make sense to pass the author if its empty, does it?

b1rger (approved):
I am approving this commit, because I think it is programmatically correct.

Nonetheless some feedback:
I prefer f-strings to `.format` - I think they are more readable.
And I think the commit messages could be a bit more verbose, but YMMV.
Regarding the reformatting of the HTML file - I think we should do this like with the black linter in a one time commit that reformats all the files. If we make a commit for every html file that clutters the commit history and makes it impossible to trace back changes.

steffres:
I agree with the html formatting done in one dedicated commit.

apis_core/apis_metainfo/models.py
koeaw: Oups, true about the author, this is what I get for copy/pasting, tsss.

koeaw:
> I wasn't suggesting to fix the other format strings- my suggestion was only to not introduce another one.

But I didn't introduce another format, I used what was already there (`.format()` in the one file, f-strings in the other), for reasons of consistency because that's more important IMHO than indiscriminate upgrades of code, even if those are theoretically preferable in terms of syntax or whatever.

Doesn't mean an entire file needs to be updated, but the most relevant/sensible unit of code should be consistent. In this case e.g. that entire method (or the entire test, in the other file). But sandwiching a single f-string between `.format()`ed strings in an if-statement is not an improvement.

You wouldn't randomly switch typefaces mid-sentence or mid-paragraph in human-readable text either because it degrades readability. (If you did, I'd consider it bad form. While it may be a non-issue for some, it makes processing of information much harder for others. It's the same for code, really.)

> I know its hard to find a balance between simply fixing a bug and cleaning everything up. I'm not sure myself what the best approach would be. My approach would probably to leave the other format string as they are and at some point take an hour or so to go through the codebase and replace all format strings in one commit.

For me it mostly comes down to (expected) time spent in a file / disproportionate mental effort required to work with a file.

E.g. I have no interest in reformatting the entirety of these two Python files because my work here is done, but I've previously cleaned up files which I found myself returning to because they needed work in several places/for different PRs. Often it's just about biting the bullet. Other times it's about spending a couple extra minutes when you can spare them to save the next person (whether a coworker or future you) time.

To change all string-formatting throughout the entire code base seems like a good idea but it's not something I personally would seriously consider doing because I couldn't justify the time investment (+ I simply don't mind individually reformatted files enough).

> For me as a reviewer it does not make it easier to review if there are changes unrelated to the actual fix in the PR.

Hm, yeah, as said it's a middle-ground solution. An extra commit for unimportant changes is really more a courtesy to anyone who ends up looking at the log. While there's obviously no point in checking the diff of the affected files _across the entire PR_ (nevermind GitHub royally messed up the visual diff in this case, lol), the relevant stuff can at least be looked at separately on a commit basis (i.e. you can choose to ignore the unimportant changes).

Neither the lazy/uncaring "put all changes in a single commit" nor the overkill "create separate PRs for formatting-only changes to individual files no-one really needs to look at anyway" extremes seem very useful.

> What are you using to reformat the html files? Maybe we can simply reuse that and create a cleanup commit + github action that checks future PRs?

I wish. But I currently reformat everything manually. |: It's primarily reducing indentation, fixing positioning of tags (HTML nesting, Python blocks), some CSS clean-up (incorrect or superfluous declarations, ordering of selectors, whatever catches my eye), removal of redundant whitespace etc. I've also added comments for documentation/easier navigation but I might have put those into separate commits for transparency, actually.

koeaw:
Btw. I wouldn't create a separate commit for something like switching the string formatting in this method here, for example. That's too small a change to justify the overhead; should go in the same commit as the functional change, IMO.

koeaw:
Removed superfluous author, updated commit messages (one was more than unclear, said view rather than template, err), left formatting as-is. Merging.

PR #112, opened 2023-04-04
apis_relations: fix RootObject import and remove TempEntityClass import

b1rger:
The RootObject resides in apis_core.apis_metainfo and should be imported
from there. The TempEntityClass is not used in this class at all.


steffres:
Agreed.

PR #114, opened 2023-04-04
apis_entities: fix update of relations when merging

b1rger:
Now that the relations are replaced by TempTriples, the former logic of
matching on the modelname of the relation does not work anymore. Instead
we use two simple update statements to TempTriple filters.

@steffres I'm assigning this to you, because according git `git blame` you wrote this for loop https://github.com/acdh-oeaw/apis-core-rdf/blob/65fa513d284130f3ef095afad98b42a490342437/apis_core/apis_entities/models.py#L320 - maybe you can shed some light on whether it contains logic this fix should also contain.

- [x] I am aware that `TempTriple` will be gone soon, but our users would like to merge entities


b1rger:
@sennierer reviewed and ACKed, thus merging

Issue #21, opened 2022-11-02
Tables aren't rendered correctly


koeaw:
**Describe the bug**
We use settings for our `django-tables2` tables that don't actually ever get applied.

See [these CSS classes](https://github.com/acdh-oeaw/apis-core-rdf/blob/5ac20837e42f6eef0241029bd8dd85ff1f6690bc/apis_core/apis_entities/tables.py#L79) added in `GenericEntitiesTable` in `apis_core/apis_entities/tables.py` and other places. `table-striped` renders even/odd table rows with different background colours for better readability; `table-hover` adds a visual effect on mouseover.

The CSS classes should get inserted e.g. here: [browsing/templates/browsing/generic_list.html](https://github.com/acdh-oeaw/apis-core-rdf/blob/main/browsing/templates/browsing/generic_list.html#L126) but aren't because the code to include them is faulty.

**Screenshots**
_What table rows currently look like_
![table_rows_A](https://user-images.githubusercontent.com/113020169/199568760-d0b649a3-ed70-425a-af4f-49ca4d3eae41.png)

_What table rows should look like (table has CSS class `table-striped`)_
![table_rows_B](https://user-images.githubusercontent.com/113020169/199569801-d7b5307a-dbbd-4243-817b-b270ad871944.png)

**Additional context**
Since we don't actually use `apis-core-rdf/browsing` in APIS Ontologies, I assume the code needs to be fixed in [apis-core](https://github.com/acdh-oeaw/apis-core) and the package updated/re-installed from PyPI for Ontologies projects for changes to take effect.

koeaw:
Related merged PR: https://github.com/acdh-oeaw/apis-core-rdf/pull/111

PR #116, opened 2023-04-05
ci: Introduce github workflow for formatting templates

b1rger:


koeaw:
Is there a way to set a preferred indentation with this tool? Currently, this seems to do more the opposite of what my goal is (it adds additional whitespace rather than removing it).

b1rger:
> Is there a way to set a preferred indentation with this tool? Currently, this seems to do more the opposite of what my goal is (it adds additional whitespace rather than removing it).

Yes, `-t` - what is your preferred indentation?

b1rger:
I've now updated the PR and set the indentation to 2 in the `pyproject.toml` - is two spaces an ok indentation?

koeaw:
Oh no, looks like the reply I'd typed out never made it into the issue. ): Trying to recreate it from memory...

> Yes, `-t` - what is your preferred indentation?

2 spaces instead of 4, and the outermost Python block shouldn't get indented because that doesn't really do anything for readability.

The main overall concern is screen real estate because lines become very long very quickly (due to HTML nesting, which isn't a necessity from a technical POV but invaluable for readability), and neither horizontal scrolling nor the results of IDEs soft-wrapping lines are great to deal with. Reason why 2 spaces for indentation have become a thing (4 used to be the default) and why certain tags – like `<html>`, `<head>`, `<body>`  – are often not indented anymore at all, because the important stuff is what goes _inside_ the `<body>` tag.

koeaw:
Look like we posted at the same time!

>  is two spaces an ok indentation?

Yes!

Does it have to be told separately for HTML, CSS and JS? Because I think I'm still seeing 4 spaces in some files.

koeaw:
I'll look at the files locally, the GitHub diff doesn't necessarily make regular indentation and lining up of tags obvious.

Did you end up switching the tool? Just wondering where I should look for documentation.

koeaw:
> Did you end up switching the tool? Just wondering where I should look for documentation.

Ok, just saw it's indeed djlint now!

koeaw:
Related issue: acdh-oeaw/apis-rdf-devops#21

koeaw:
> Is the linting fine the way it is for now? Should we change some configuration?

Sorry, I haven't got around to actually trying it out so far. Will try to make time this week!

koeaw:
@b1rger I ended up installing the package to work around the merge conflicts & to take a closer look at its capabilities in general because it would be preferable if it could be used with `apis-rdf-devops` / for local development (i.e. #21).

I'm thinking it might make sense to move further discussion to the other issue.

b1rger:
Now that we decided to use poetry for the github workflows, this is blocked by #214

PR #120, opened 2023-04-05
move AdditionalSerializerConfig to core so that it can be reused in o…

steffres:


b1rger:
You can can change existing commits using `--amend` - you don't have to create an extra commit for stylistic fixes as long as its not merged

steffres:
> You can can change existing commits using `--amend` - you don't have to create an extra commit for stylistic fixes as long as its not merged

fair enough.

PR #121, opened 2023-04-06
run pycharm's import optimizer

steffres:
this does:
* sorts and groups imports according to pep8
* removes unused imports


koeaw:
What about all the commented import statements? I saw some of them were moved to different lines, but shouldn't they also go?

koeaw:
> But this is out of scope of this PR as this is just an automated clean-up.

I see. Those commented imports being scattered across the import blocks aside, the clean-up of active statements seems to not have been done properly by PyCharm, either, unfortunately – it didn't sort everything + added random blank lines.

I may have had this happen to me as well but don't know if/how I managed to solve it. Possibly be running the optimization action several times.

steffres:
What was not sorted correctly?

And the blank lines are in accordance with pep8 on grouping imports: https://peps.python.org/pep-0008/#imports

koeaw:
> And the blank lines are in accordance with pep8 on grouping imports: https://peps.python.org/pep-0008/#imports

Yep, was just going to amend my comment to say the lines may actually be due to grouping!

Will try to find examples for the sorting problem, hold on.

steffres:
You mean alphabetic sorting? I'm not fond of it, as changes in imports would require re-sorting, and the imports are hardly read anyway.

Feel free to sort it though if you're inclined.

koeaw:
Well, it's PyCharm's default behaviour. Personal preferences are fine, but references to IDE-specific tools or functionality will make fellow devs assume they are used the usual way, not with unknown custom settings (unless explicitly noted).

PR #123, opened 2023-04-06
chore: update to 3.11

b1rger:


.github/workflows/django-tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8"]
+        python-version: ["3.11"]
```
sennierer, line 16: why not keep test for all versions between 3.8 and 3.11? or at least 3.10 and 3.11

b1rger: fixed

apis/urls.py
```diff
@@ -1,70 +1,78 @@
...
 from django.contrib import admin
 from django.urls import path
 
 from apis_core.apis_entities.api_views import GetEntityGeneric
 
 if "theme" in settings.INSTALLED_APPS:
     urlpatterns = [
-        url(r"^apis/", include("apis_core.urls", namespace="apis")),
-        url(r"^api-auth/", include("rest_framework.urls", namespace="rest_framework")),
+        re_path(r"^apis/", include("apis_core.urls", namespace="apis")),
```
sennierer, line 10: we should use `re_path` only when regex is really needed. for normal routes `path` is enough. eg `path("apis/", include("apis_core.urls", namespace="apis")),`

apis_core/apis_labels/urls.py
```diff
@@ -1,13 +1,15 @@
...
 app_name = "apis_labels"
 
 urlpatterns = [
-    url(r"^list/$", views.LabelListView.as_view(), name="label_list"),
-    url(r"^create/$", views.label_create, name="label_create"),
-    url(r"^edit/(?P<pk>[0-9]+)$", views.label_edit, name="label_edit"),
-    url(r"^delete/(?P<pk>[0-9]+)$", views.LabelDelete.as_view(), name="label_delete"),
+    path("list/", views.LabelListView.as_view(), name="label_list"),
+    path("create/", views.label_create, name="label_create"),
+    re_path(r"^edit/(?P<pk>[0-9]+)$", views.label_edit, name="label_edit"),
```
sennierer, line 11: this can use path as well `path(edit/<int:pk>/...`

sennierer (changes requested):
please switch  `re_path` for `path` where possible

b1rger:
> please switch `re_path` for `path` where possible

done

sennierer (approved):
perfect, thanks!

sennierer:
can be merged from my point of view

PR #125, opened 2023-04-11
style: use pythons datetime instead of reinventing the wheel

b1rger:
This also moves the nested function outside of `parse_date` - there is
no reason to nest it

### DIFF
--- a/apis_core/helper_functions/DateParser.py
+++ b/apis_core/helper_functions/DateParser.py
@@ -3,6 +3,18 @@ import re
 from datetime import datetime, timedelta
 
 
+def get_last_day_of_month(month, year):
+    """
+    Helper function to return the last day of a given month and year (respecting leap years)
+
+    :param month : int
+    :param year : int
+    :return day : int
+    """
+    next_month = datetime(year=year, month=month, day=28) + timedelta(days=4)
+    return (next_month - timedelta(next_month.day)).day
+
+
 def parse_date(date_string: str) -> (datetime, datetime, datetime):
     """
     function to parse a string date field of an entity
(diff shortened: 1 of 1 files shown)


steffres:
I see my implementation as more readable. It's a simple and stable formula, taken straight from the calendar definition, that describes itself. Your implementation requires some mental hoops to jump through (create a datetime object, add days to it so that it's the next month, subtract from it its own days as timedelta, then pick that day). This a curvy logical route compared to my transparent branching conditionals.

On the nesting, that is on purpose to keep the highest scope of the module clean. As with variables, if a function is only used within a local scope it's better for it to stay there; it has no business being global in the highest scope.

koeaw:
I personally like the proposed solution a lot more – it's clear and concise yet not minimal in a "bad" way (doesn't use obscure functionality nor is overly complex). Very Zen of Python. (: It took me a moment to work out the return statement but that was due to an incorrect or incomplete assumption about the value of the variable on my part.

The only reason for keeping the current solution would be speed (if applicable), IMO.

I didn't look at or into the containing code but think a general purpose function is best made available for independent consumption. I'm assuming the date parsing is currently used for the fields we're dropping (by dropping `TempEntityClass`)? I don't necessarily see everyone continuing to use that function but making generic parts of it available for reuse seems like a good idea.

sennierer:
I also think that the proposed solution is much cleaner than the existing: as mentioned the datetime module is widely adopted (and the function even returns datetime objects), so no need to reimplement already existing solutions.
I think speed is not an issue here, as the function is mainly used for parsing manually entered data.

koeaw:
Hmm, I just assigned myself as a reviewer only because the GitHub web UI suggested that's what the PR was waiting for (was assuming it wanted the green :heavy_check_mark:), but now it says it doesn't. :thinking:

koeaw (approved):

b1rger:
This PR was only a [draft](https://dict.leo.org/englisch-deutsch/draft). I'm not sure myself what the best way to improve the code is, I'm still playing with ideas. When I think the PR is ready for review, I'll push the `Ready for review` button.

(I'll stop creating draft PRs for now, apparently the `draft` status is too easy to be overlooked and it leads to people wasting time with reviews of WIP code)

PR #126, opened 2023-04-12
fix: use logger instead of an exception

b1rger:
Even if there is an unusable filter, doesn't mean the remaining
interface could not still be used. Using logging we can inform the admin
of the misconfiguration, but still show the form.

### DIFF
--- a/apis_core/apis_entities/filters.py
+++ b/apis_core/apis_entities/filters.py
@@ -1,3 +1,5 @@
+import logging
+
 from collections import OrderedDict
 
 import django_filters
@@ -126,12 +128,10 @@ class GenericEntityListFilter(django_filters.FilterSet):
                         default_filters[field_name].label = filter_settings["label"]
 
                     field_filters[field_name] = default_filters[field_name]
-
             else:
-                raise ValueError(
-                    f"Filters for individual entities need to be of type "
-                    f"string or dictionary.\n"
-                    f"Got instead: {type(f)}"
+                logging.error(
+                    "Unusable filter - should be name of a default filter or a dict, got %s",
+                    f,
                 )
 
         for f_name, f_filter in default_filters.items():


steffres:
No, this is a dedicated exception which tells you what went wrong.

koeaw:
I'm confused, does this mean you've changed your opinion? Two months ago, you agreed with what I said...

koeaw:
Or, to be more precise, at least acknowledged the argument/didn't disagree.

steffres:
I agreed with you that Django breaking by itself is not helpful often, as it might not tell you what broke.

But this exception here is a custom one which tells you exactly what broke. That is helpful.

koeaw:
Ok, that wasn't obvious to me at all as I specifically singled out this very exception as being unhelpful; it's what the bulk of my comment was about. My reasoning remains the same, so I guess we'll have to agree to disagree on what is (or isn't) helpful.

b1rger:
> But this exception here is a custom one which tells you exactly what broke. That is helpful.

I'not sure how this exception is helpful, because, ehm, a string is a string. But YMMV:

> Filters for individual entities need to be of type string or dictionary.
> Got instead: <class 'str'>

But whatever the Exception or the Logger says: with a9eb86831c1f180a52b633668b06e09a9afe1c7d we fixed the entities settings, so that they are read from the settings file again (as discussed). If projects have a `settings.py` containing entity settings (like for example the example project) and their filter definitions contain strings that does not adhere to the expectations of the `set_sort_filters` method (like for example in the example project) the projects would now get an exception. Just by updating. In the current setup, where we have *no* changelog and *no* release management, the most sensible thing to do is to try *not* to break existing installations.

steffres:
>  the projects would now get an exception. Just by updating.

Ah, I missed that. How so? And what does updating mean in this context? Pushing a new commit to apis-ontologies?

Issue #133, opened 2023-04-13
Find and get rid of `webpage` dependencies


b1rger:
`webpage` is a frontend, that is part of another repository. some of the functionality (mostly the `base.html` template, but also some variables) are used throughout the `apis_core` codebase. Those references to `webpage` should be removed / replaced, so that `apis-rdf-core` can function without `webpage`

PR #134, opened 2023-04-13
refactor: get rid of `webpage` dependency

b1rger:
`webpage` is a frontend that is part of another repository. some of the
functionality (mostly the `base.html` template, but also some variables)
were used throughout the apis_core codebase. The `base.html` was
replaced by a thinned down `base.html` in the apis templates folder.
Another template that was used from `webpage` was the
`confirm_delete.html` template, merged with the same template in
`apis_entities` and moved to the main `templates` folder.
Some views relied on some webpage functionality, those code snippets
were replaced.
A `shared_url` context_processor was also copied from `webpage` and
enhanced.
Closes: #133

### DIFF
--- a/apis_core/apis_entities/edit_generic.py
+++ b/apis_core/apis_entities/edit_generic.py
@@ -277,7 +277,7 @@ class GenericEntitiesDeleteView(DeleteView):
     #     app_label='apis_entities', model='tempentityclass').model_class()
     model = importlib.import_module("apis_core.apis_entities.models").TempEntityClass
     template_name = getattr(
-        settings, "APIS_DELETE_VIEW_TEMPLATE", "apis_entities/confirm_delete.html"
+        settings, "APIS_DELETE_VIEW_TEMPLATE", "confirm_delete.html"
     )
 
     def dispatch(self, request, *args, **kwargs):
--- a/apis_core/apis_metainfo/templates/base.html
+++ b/apis_core/apis_metainfo/templates/base.html
@@ -1,0 +1,203 @@
+<!DOCTYPE html>
+{% load static %}
+{% load browsing_extras %}
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>{% block title %}Title Placeholder{% endblock %}</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  {% block metaDescription %}
+    <meta name="description" content="Description">
+    <meta name="author" content="Author">
+  {% endblock %}
+  <!-- Start favicons -->
+  <link rel="apple-touch-icon" sizes="180x180" href="{{ SHARED_URL }}favicon/apple-touch-icon.png"/>
+  <link rel="icon" type="image/png" sizes="32x32" href="{{ SHARED_URL }}favicon/favicon-32x32.png"/>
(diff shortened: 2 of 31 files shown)


apis_core/apis_entities/edit_generic.py
```diff
@@ -277,7 +277,7 @@ class GenericEntitiesDeleteView(DeleteView):
     #     app_label='apis_entities', model='tempentityclass').model_class()
     model = importlib.import_module("apis_core.apis_entities.models").TempEntityClass
     template_name = getattr(
-        settings, "APIS_DELETE_VIEW_TEMPLATE", "apis_entities/confirm_delete.html"
```
sennierer, line 280: is that correct? was the confirm delete also moved to root?

b1rger: Yes, there was already one `confirm_delete.html` template in `apis_entities` and the one shipped with `webpage`. They were identical, I moved one of them to the root template folder of `apis_metainfo` and deleted the one in `apis_entities`

apis_core/apis_metainfo/templates/base.html
sennierer, line 1: is that the `base.html` template mentioned everywhere? Wouldnt that end up in `apis_core/apis_metainfo/base.html` rather than `base.html`?

sennierer (approved):
if the new `base.html` template really ends up at root template folder this looks good to me

koeaw:
Question: where should modifications to `base.html` be applied in future? Both in `apis-rdf-devops`  (for now; not sure how these changes affect https://github.com/acdh-oeaw/apis-rdf-devops/issues/59) and `apis_metainfo`?

b1rger:
> Question: where should modifications to `base.html` be applied in future? Both in `apis-rdf-devops` (for now; not sure how these changes affect [acdh-oeaw/apis-rdf-devops#59](https://github.com/acdh-oeaw/apis-rdf-devops/issues/59)) and `apis_metainfo`?

That depends on the change. The `base.html` shipped with `apis-core` is a simple default template without any project or org specific code. The `webpage` app is an org specific frontend. If the latter is installed, its `base.html` overrides the one from apis-core.

b1rger:
> if the new `base.html` template really ends up at root template folder this looks good to me

I tested it one more time and there is also only one `base.html` in this codebase.

b1rger:
> That depends on the change. The `base.html` shipped with `apis-core` is a simple default template without any project or org specific code. The `webpage` app is an org specific frontend. If the latter is installed, its `base.html` overrides the one from apis-core.

I realized that this is not entirely correct - given the `webpage` module has its template in `webpage/base.html`, the inheritance does not use that one. We do now have multiple options:

* *also* change the `base.html` path in the `webpage` module so that it overrides the `base.html` from `apis-core`
* move the `base.html` from `apis-core` back to a `webpage` folder (I think thats technical debt)
* use a setting that points to the correct `base.html`

I would prefer the first option

b1rger:
> I would prefer the first option

I found an solution that is a combination of option 1 and 3: I added a value to the `custom_context_processors.list_apis_settings` dict that sets `basetemplate` either to `settings.BASE_TEMPLATE` or to `base.html` as a fallback. I've also updated the `apis/settings/base.py` settings file accordingly - its `BASE_TEMPLATE` now points to `webpage/base.html`

PR #137, opened 2023-04-17
style: Use Mixin instead of decorator in views

b1rger:


koeaw (approved):
The only things I noticed which would benefit from adaptation lie outside the scope of this PR (`success_url` on deletion would ideally respect sort order of Uris; list view QuerySet isn't sorted, which prompts Django to issue a warning).

b1rger:
Thanks for the review!

> The only things I noticed which would benefit from adaptation lie outside the scope of this PR (`success_url` on deletion would ideally respect sort order of Uris; list view QuerySet isn't sorted, which prompts Django to issue a warning).

I'll try to keep that in mind when working on #138

Issue #140, opened 2023-04-19
apis_metainfo: add `merge` skeleton method to `TempEntityClass`


b1rger:
Now that we are using inheritance extensively, we should add a skeleton method for merging objects to `TempEntityClass` - this method would then be called by the `merge_with` method. (Another approach would be to overload the `merge_with` method and use `super()`, but that method deletes the merged object in the end, so that approach would not work...)

PR #142, opened 2023-04-19
feat(apis_entities): copy fields when merging entities

b1rger:
This commit introduces a `merge_fields` method in the TempEntityClass.
It copies information for the relevant fields in TempEntityClass and can
be extended by subclasses to implement custom logic for copying custom
subclass fields.

Closes: #140

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -275,6 +275,52 @@ class TempEntityClass(AbstractEntity):
             kwargs={"entity": entity, "pk": self.id},
         )
 
+    def merge_charfield(self, other, field):
+        print(field)
+        res = getattr(self, field.name)
+        if not field.choices:
+            if getattr(other, field.name):
+                res += " (" + getattr(other, field.name) + ")"
+        setattr(self, field.name, res)
+
+    def merge_textfield(self, other, field):
+        res = getattr(self, field.name)
+        if getattr(other, field.name):
+            res += "\n" + "Merged from {other}\n" + getattr(other, field.name)
+        setattr(self, field.name, res)
+
+    def merge_booleanfield(self, other, field):
+        setattr(
+            self, field.name, getattr(self, field.name) and getattr(other, field.name)
+        )
+
+    def merge_start_date_written(self, other):
+        self.start_date_written = self.start_date_written or other.start_date_written
+
+    def merge_end_date_written(self, other):
+        self.end_date_written = self.end_date_written or other.end_date_written
+
+    def merge_fields(self, other):
+        """
+        This method iterates through the model fields and copies
+        data from other to self. It first tries to find a merge method
+        that is specific to that field (merge_{fieldname}) and then tries
+        to find a method that is specific to the type of the field (merge_{fieldtype})
+        It is called by the `merge_with` method.
+        """
+        for field in self._meta.fields:
+            fieldtype = field.get_internal_type().lower()
+            # if there is a `merge_{fieldname}` method in this model, use that one
+            if callable(getattr(self, f"merge_{field.name}", None)):
+                getattr(self, f"merge_{field.name}")(other)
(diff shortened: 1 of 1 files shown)


b1rger:
Please note that I made some decisions about how attributes are merged- i.e. the `name` field gets concatenated with a comma. There might by better approaches.

koeaw:
@b1rger I'm not entirely clear on the long(er)-term purpose of this seeing as we're about to drop `TempEntityClass`. Is this supposed to be moved elsewhere later on? Or can this help with _moving away_ from `TempEntityClass`?

I had a look at the linked issue but am none the wiser. It's possible I'm only missing a bit of context here.

b1rger:
> @b1rger I'm not entirely clear on the long(er)-term purpose of this seeing as we're about to drop `TempEntityClass`. Is this supposed to be moved elsewhere later on? Or can this help with _moving away_ from `TempEntityClass`?
> 
> I had a look at the linked issue but am none the wiser. It's possible I'm only missing a bit of context here.

At the moment when entities are merged information gets lost and there is no way for ontologies to define how the local fields of their entities should be handled when merging is done. I don't believe that `TempEntityClass` will be going away that soon, so I think it makes sense to fix this problem

apis_core/apis_entities/models.py
```diff
@@ -274,6 +274,26 @@ def get_duplicate_url(self):
...
         )
 
+    def merge_fields(self, other):
+        """
+        This method is intended to be extended by subclasses.
+        It is called by the `merge_with` method and should transfer data from `other` to `self`
+        """
+        # Inherited from RootObject
+        if self.name != other.name:
+            self.name = ", ".join([self.name, other.name])
```
sennierer, line 284: I think it should be rather
```python
self.name = f"{self.name} ({other.name})"
```
for the dates `self` is overruling `other` when both objects have dates, that should be reflected in the string based fields as well

apis_core/apis_entities/models.py
```diff
@@ -274,6 +274,26 @@ def get_duplicate_url(self):
...
+        # Boolean values
+        self.review = self.review and other.review
+        self.status = self.status and other.status
+        self.published = self.published and other.published
+        # date strings: we use the other date only if the self date is missing
+        self.start_date_written = self.start_date_written or other.start_date_written
+        self.end_date_written = self.end_date_written or other.end_date_written
+        # we simply concatenate strings
+        self.references = "\n".join(filter(None, [self.references, other.references]))
```
sennierer, line 293: see comment above. should be clear which comments stem from an object that was merged in the actual one (put in brackets or explicitly mentioned ... eg `references from merged objects:`

sennierer (changes requested):
I think the merging is currently not completely concise:
`self` is overruling `other` on some fields, others are completely equal.
I think the object that something is merged into (so `self`) should be overruling `other` and this should also be reflected in the string based fields.

Given that this should work on all classes, so include also fields that are not inheritated from `RootObject` we could refactor that to run through all fields and depending on the field type (so whether its sting, date etc. always use the same logic/function on it.

b1rger:
> Given that this should work on all classes, so include also fields that are not inheritated from RootObject we could refactor that to run through all fields and depending on the field type (so whether its sting, date etc. always use the same logic/function on it.

Oke, I've implemented that - the `merge_fields` method now iterates through the model fields and checks wether there is a `merge_{fieldname}` method and if so uses that. If not, it checks for a `merge_{fieldtype}` method and uses that. If no method exists and the field in `self` is empty, it copies the value of `other`.

I've implemented `merge_` methods for charfield, textfield and booleanfield. Given that `start_date_written` and `end_date_written` are parsed and the parsing would fail with the default merge method for charfields, I added extra methods for those two, demonstrating how the merge methods get chosen.

In the long run, some of the merge functionality should probably move to the RootObject or the AbstractEntity...

PR #147, opened 2023-04-27
chore: update charts dependency to 0.6

b1rger:
This fixes a warning in Django versions >= 3.2, see
https://github.com/acdh-oeaw/acdh-django-charts/commit/324270aed90ad446468329cd1d0bbf64127d873c


b1rger:
already fixed in #146

Issue #149, opened 2023-05-02
Tables can't be scrolled horizontally


koeaw:
**Describe the bug**
List view tables with lots of columns get cut off when they become too wide.

**To Reproduce**
1. Add field names for extra columns to `additional_cols` to an entity's `APIS_ENTITIES` settings.
2. While browsing the entity, select a number of additional columns to be added to it from the dropdown on the left.
3. Check your table. (In case you use an extra-wide screen, resize your browser window to simulate a narrower window.)

**Expected behavior**
_All_ columns should be visible/sortable/...

**Screenshots**
![Screenshot from 2023-05-02 18-59-26](https://user-images.githubusercontent.com/113020169/235734584-e7c34a80-6f96-436e-b1dc-44942ad5d278.png)

PR #150, opened 2023-05-02
Fix table rendering

koeaw:
**Describe your changes**

* Fixes how the CSS for tables from `tables.py` is included in `generic_list.html` so it actually works -> tables are supposed to have striped rows and the row being hovered over should be highlighted.
* Surrounds the table with a `div` to make it responsive -> allows horizontal scrolling of wide tables instead of cutting off columns.

**Related issues and PRs**

Resolves issues:
- #21 
- #149

### DIFF
--- a/apis_core/apis_metainfo/templates/browsing/generic_list.html
+++ b/apis_core/apis_metainfo/templates/browsing/generic_list.html
@@ -116,7 +116,8 @@
                                 </div>
                             {% endif %}
                         {% endwith %}
-                        {% block table %}
+                        <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


katharinawuensche:
Sorry that this took me so long, I was having trouble with tables on my local instance since the Postgres experiments which had to be fixed first for the review. The changes work like a charm (tested in Chrome, Firefox and Edge)!

katharinawuensche (approved):

koeaw:
Thx for the review! Just merged this. :tada:

PR #154, opened 2023-05-10
feat: apis_entities: introduce apis_entities.mixins

b1rger:
This commit introduces the EntityMixin and the EntityInstanceMixin which
can be used in views that handle entities. This way all those views have
access to the same class attributes.


gregorpirgie (approved):
Looks good, I support this!

PR #157, opened 2023-05-17
Refactor RDF Parser

b1rger:
This PR introduces a new RDF Parser logic. The first commit introduces the
basic helper functions, the second commit implements them in the model and the
third commit adds a `getorcreate` view for the URIs.

- feat: implement new rdf parser
- fix: harden and enhance Uri model
- feat: Implement form, view and route for adding URIs


b1rger:
We should probably update the `RDF_default_settings.yml` for the new parser

sennierer:
I think the yaml parser library and/or the toml parser library prior v 3.11 are very simple libraries that we can safely rely on. If the built in toml parser is the only reason to set python to >= 3.11 I would go with the parser libraries and less strict python requirements.

apis_core/apis_entities/api_views.py
```diff
@@ -395,8 +394,9 @@ class GetOrCreateEntity(APIView):
     def get(self, request):
         entity = request.query_params.get("entity2", None)
         uri = request.query_params.get("uri", None)
-        if uri.startswith("http:"):
-            ent = RDFParser(uri, entity.title()).get_or_create()
+        if uri and uri.startswith("http:"):
```
sennierer, line 397: this could be `https` as well

apis_core/apis_entities/management/commands/normalize_uris.py
```diff
@@ -22,10 +16,11 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         for x in Uri.objects.all():
```
sennierer, line 17: I think this whole block should be wrapped in `with transaction.atomic():` that way URIs get only changed if the whole set of changes succeeds.

sennierer (approved):
Added only one comment on the management command for cleaning URIs. Rest looks great to me.

PR #160, opened 2023-05-23
fix(utils): fix bug and add test case

b1rger:
The `get_all_contenttype_classes` method checked for abstact classes by
looking at the `_meta` attribute without checking if that `_meta`
attribute even existed. This commit adds an additional check and adds a
testcase.

### DIFF
--- a/apis_core/utils/test_caching.py
+++ b/apis_core/utils/test_caching.py
@@ -1,0 +1,35 @@
+from django.test import TestCase
+
+from apis_core.utils import caching
+
+all_class_modules_and_names = {
+    "apis_metainfo": [
+        "Collection",
+        "RootObject",
+        "Source",
+        "Text",
+        "Uri",
+        "UriCandidate",
+    ],
+    "apis_vocabularies": [
+        "CollectionType",
+        "LabelType",
+        "TextType",
+        "VocabNames",
+        "VocabsBaseClass",
+        "VocabsUri",
+    ],
+    "apis_entities": ["TempEntityClass"],
+    "apis_relations": ["TempTriple", "Property", "Triple"],
+    "contenttypes": ["ContentType"],
+}
+
+
+class CachingTest(TestCase):
+    def test_get_all_class_modules_and_names(self):
+        expected = []
+        for key in all_class_modules_and_names:
+            for name in all_class_modules_and_names[key]:
+                expected.append((key, name))
+        res = caching.get_all_class_modules_and_names()
+        self.assertEqual(res, expected)
--- a/apis_core/utils/caching.py
+++ b/apis_core/utils/caching.py
@@ -316,12 +316,13 @@ def get_all_contenttype_classes():
                 if not cls_n.startswith("__") and "__module__" in list(
                     dir(getattr(m, cls_n))
                 ):
+                    mod = getattr(m, cls_n)
                     if (
-                        getattr(m, cls_n).__module__ in apis_modules
-                        and getattr(m, cls_n) not in lst_cont
+                        mod.__module__ in apis_modules
+                        and mod not in lst_cont
                         and cls_n.lower() not in models_exclude
-                        and inspect.isclass(getattr(m, cls_n))
-                        and getattr(m, cls_n)._meta.abstract is False
+                        and inspect.isclass(mod)
+                        and (hasattr(mod, "_meta") and mod._meta.abstract is False)
                     ):
                         lst_cont.append(getattr(m, cls_n))
         lst_cont.append(ContentType)


koeaw:
Hmm, I believe the check for `._meta.abstract` happens at a point when we've already discarded everything that isn't a model class, and every model class has `._meta.abstract` set to something, whether or not there is a user-provided `Meta` class (much like `._meta.verbose_name` gets set for every model class).

... Having had a closer look at what this all does after getting a bit lost (again) in the many things we seem to be checking for here: the point of this function is basically to return all existing `ContentTypes`, i.e. the contents of the database table which holds them, or at least a subset thereof, is that right? (And `ContentTypes` correspond to model classes?)

b1rger:
> [...] every model class has `._meta.abstract` set to something [...]

maybe every model, but not every class. If I add
```
from django.contrib import admin
class PersonAdmin(admin.ModelAdmin)

admin.site.register(Person, PersonAdmin)
```
to my ontologies `model.py`, I get: `AttributeError: type object 'PersonAdmin' has no attribute '_meta'`

(The same happens if I try to use the [django reversion admin integration](https://django-reversion.readthedocs.io/en/stable/))


> ... Having had a closer look at what this all does after getting a bit lost (again) in the many things we seem to be checking for here: the point of this function is basically to return all existing `ContentTypes`, i.e. the contents of the database table which holds them, or at least a subset thereof, is that right? (And `ContentTypes` correspond to model classes?)

I'm not sure- the code is a mess. I wrote the tests based on what the function returns at the moment, but I plan to refactor the whole thing at some point (or throw it away and use ContentType.objects.all() or someting similar...). But I haven't had time to get an overview where it is used.

koeaw:
> maybe every model, but not every class. If I add
> 
> ```
> from django.contrib import admin
> class PersonAdmin(admin.ModelAdmin)
> ```
> (snip)

Ah, yeah, makes sense.

I thought `models.py` was only supposed to contain models – but that may be an incorrect assumption on my part, or may have been the idea so far but is not how we want to or can continue doing things. ... I feel like this has come up before during discussions of mixins? I.e. stuff in `apis_core` is currently not set up to deal with non-model classes in `models.py` or something.

> I'm not sure- the code is a mess. I wrote the tests based on what the function returns at the moment, but I plan to refactor the whole thing at some point (or throw it away and use ContentType.objects.all() or someting similar...). But I haven't had time to get an overview where it is used.

Yeah, I was wondering if it's just about `ContentTypes`... if we couldn't just replace all this super complicated/complex checking we're doing here with something simple like:

```
from django.apps import apps

(...)

content_types = []

apis_apps = [a for a in apps.app_configs.keys() if a.startswith("apis_")]

for ct in ContentType.objects.filter(app_label__in=apis_apps):
    m = ct.model_class()
    content_types.append(m)

_contenttype_classes = content_types
```

b1rger:
> if we couldn't just replace all this super complicated/complex checking we're doing here with something simple like

I would propose not to hardcode application name prefixes. Depending on how the function is used, it might make sense to simply list all children of `AbstractEntity` and `AbstractRelation`. But that needs more research.

For now it would be great if we could simply fix that bug, I stumbled over this in two occations in the last week.

apis_core/utils/test_caching.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class CachingTest(TestCase):
+    def test_get_all_class_modules_and_names(self):
+        expected = []
+        for key in all_class_modules_and_names:
+            for name in all_class_modules_and_names[key]:
+                expected.append((key, name))
+        res = caching.get_all_class_modules_and_names()
+        self.assertEqual(res, expected)
```
koeaw, line 35: As said, I'm unable to run the test at this point, what I'm wondering about is:

Wouldn't  `caching.get_all_class_modules_and_names()` include all `apis_ontology.models` as well?

(Or... is that the problem with my testing the test? Is this supposed to be limited to apis-core, i.e. not actually meant up to pick up the ontology models?)

koeaw (approved):

b1rger:
> > For now it would be great if we could simply fix that bug, I stumbled over this in two occations in the last week.
> 
> Yup, was going to approve the PR anyway. – Despite not being able to run/test the test! /:
> 
> Dunno why but I can't get Django to pick up the test. Then realised I don't even know what testing tool is being used when I run the tests from the CLI (... which is related to the general Q of what's supposed to go in our dev packages/an answer I still owe in another ticket). Then I wasn't able to get testing to work at all in the IDE. ¯_(ツ)_/¯

```
$ ./manage.py test
```
https://docs.djangoproject.com/en/4.1/topics/testing/overview/#running-tests

apis_core/utils/test_caching.py
b1rger: the tests are testing the `apis-core-rdf` project. there is no `apis_ontology.models` module anywhere.

b1rger: (In fact, the tests break when used with `apis_ontology`, because `apis_ontology` is not a real django app (yet))

koeaw:
> ```
> $ ./manage.py test
> ```
> 
> https://docs.djangoproject.com/en/4.1/topics/testing/overview/#running-tests

Spent way too much time reading the docs, various SO posts & more yday trying to figure this out. Existing tests in `apis_core`, `apis_core.utils` are being recognised/can be drilled down to. Probably a case of PEBCAK (: but will have to revisit this another time.

PR #162, opened 2023-05-23
feat(apis_entities): add additional blocks to generic detail template

b1rger:
This way we can add custom content to the entity header

### DIFF
--- a/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
+++ b/apis_core/apis_entities/templates/apis_entities/detail_views/detail_generic.html
@@ -38,6 +38,7 @@
               <small>{{ entity_type }}s</small>
             {% endif %}
             {{ object }}
+            {% block object-actions %}
             {% if user.is_authenticated %}
               {% if object.get_edit_url %}
                   <a href="{{ object.get_edit_url }}">
@@ -66,6 +67,8 @@
                   </a>
               {% endif %}
             {% endif %}
+            {% endblock object-actions %}
+            {% block object-actions-extra %}{% endblock object-actions-extra %}
           </h2>
         </div>
       <div class="col-md-2">


koeaw (commented):
I'm not exactly sure what these blocks are supposed to be used for.

I added a `<div>` in each place to try to better understand, which results in the "main" text (/the `<h2>` headline) having more lines added to it:

Screenshot 2023-05-25 at 11-07-53 Alles Wolken (ID 639)

Are you already using this or is there a WIP thing for which you have a visual you could share?

As a general note, it's always best to avoid naming frontend elements using positioning (/colour/size/formatting,... any styling specifics) because the names quickly lose meaning with responsive layouts or aren't "built to last".

Btw. I know the styling is ugly and messy so less than ideal to work with. I might even have started to clean that particular area up the littlest bit in a local branch or stash before, which I could add onto your PR (or follow it up) if I find it. Just need to know what the blocks are meant for to be able to better help.

koeaw:
Ahh, more icons, gotcha.

And what's the `auth` block for? Sounded to me like maybe for additional login functionality or similar.

> > As a general note, it's always best to avoid naming frontend elements using positioning (/colour/size/formatting,... any styling specifics) because the names quickly lose meaning with responsive layouts or aren't "built to last".
> 
> Ah, I didn't know that - I'm open for suggestions. Maybe `entity-menu-after`?

`after` has the same problem (positioning). ^^ Better to use sth. to do with purpose.

I dunno what the new icon does or potentially other new ones would be used for, but all of them taken together – i.e. not just additional ones – could probably be described as... `entity-actions`? So, thinking ahead, that might make a good name for a block (whether a containing HTML element or Django template block or both) to group _all of them_ together. Following on from that, maybe sth. like `entity-actions-extra` for an optional block contained within this (future) main actions block?

What about icon order? Are they strictly meant to appear after existing ones? (E.g. I already don't like that the delete icon sits between create and duplicate instead of being last.) Is there a possibility that optional icons should rather come first?

A refactor of the current template will separate out the 3 main concerns here (leaving aside the next/previous arrows) – the clickable entity type, the name/ID of the current entity object, the set of action icons/buttons – by putting them into 3 independent HTML containers for more flexibility wrt layout. The only element designated a headline would/should then be the name of the entity being viewed (which will also affect the current separator between the icons).

koeaw:
Or maybe... `entity-object-actions` + `-extra` for less ambiguous wording. We often don't differentiate clearly between entities and entity objects.

b1rger:
> Ahh, more icons, gotcha.
> 
> And what's the `auth` block for? Sounded to me like maybe for additional login functionality or similar.

The icon is for displaying the bibsonomy references. The `auth-` block was meant as an additional block that could contain icons that allow actions for authenticated users (because there is a `{% is user.is_authenticated %}` block above) - but that might be nonsense, given that the whole interface is only accessible for authenticated users anyway.

b1rger:
> Or maybe... `entity-object-actions` + `-extra` for less ambiguous wording. We often don't differentiate clearly between entities and entity objects.

I've now called it `object-actions-extra` and also put the existing icons above in an `object-actions` block. I chose not to use the `entity` term, because we might want to reuse that code as a partial at other places

koeaw (approved):

koeaw:
> I've now called it `object-actions-extra` 

Sounds good!

koeaw:
In instances where different user groups see slightly different things depending on their permissions, esp. when it's  interactive elements like a row of buttons or links, it's always best to order them in the least disruptive way. E.g. to put elements with stricter permissions after ones with lesser ones, so elements don't jump around the page/switch positions when logging in/out etc., which could lead to accidental wrong clicks.

Not applicable here, but needs to be taken into consideration in other situations. Might also be part of why I find the ordering of the existing icons irritating (with multiple permission levels, the trash icon wouldn't necessarily be available to everyone, so should be last).

Issue #163, opened 2023-05-26
Fix `(templates.E003) 'browsing_extras' is used for multiple template tag modules`


b1rger:
Django warns at startup:
```
?: (templates.E003) 'browsing_extras' is used for multiple template tag modules: 'apis_core.apis_metainfo.templatetags.browsing_extras', 'webpage.templatetags.browsing_extras'
```

PR #164, opened 2023-05-26
fix(apis_metainfo): rename templatetags

b1rger:
The browsing_extras templatetags were initially merged from the browsing
application to make apis-core not depend on browsing. The webpage
application also has browsing_extras templatetags and therefore the
identical names led to a `templates.E003` warning from Django.
This commit renames the templatetags module to `apis_metainfo_extras`.
It also moves the used templattag templates to the
templates/apis_metainfo/tags folder.
The commit also drops the `nav_menu` templatetag, because it turned out
that it was not used anywhere.

Closes: #163


koeaw:
Not a review, just wanted to point out a typo in the subject (assuming the subject becomes part of the log when commits are squashed) -> templatetags is missing an L.

koeaw:
> Not a review, just wanted to point out a typo in the subject (assuming the subject becomes part of the log when commits are squashed) -> templatetags is missing an L.

Uhm, dunno how, but I seem to have got responsibility for commit & PR subject confused in my head _while talking about how they interact_ -> fixed the typo myself, sry for the noise.

PR #177, opened 2023-05-31
chore!: bump Python dependency to >=3.11

b1rger:
Closes: #161


sennierer (approved):
We might want to update the lock file as well. Most probably nothing will change in the version resolution, but there could be dependencies that need other versions for Python >=3.11.

Issue #15, opened 2022-09-29
Discuss and consider removing apis-labels


SteffRhes:
Unless I overlooked it, I don't see a compelling reason in having a django app apis-labels just for a single use-case. This is ontology-specifics hard-coded in business logic. 

We might be able to replace it with semantic equivalences within apis-ontology module.

further steps from now on:
* discuss this with team at JF

b1rger:
Our JFX protocol says:
```
    Merge-function should be function of rootObject and can be overwritten in ontology
    That allows us to remove apis-labels 
```

What is the connection between the former and the latter? Can't we simply remove `apis-labels`? @sennierer maybe you know what the intention of that second line was...

sennierer:
Second line meant that if we allow for overwriting the merge functionality in ontologies, the ontologies can also take care of storing whatever information they want to keep in whatever model they want to use by adding a pre-delete signal and checking for  the issuer of that signal.
As the one of the main  functions in apis-rdf of the labels model was to keep information (such as the label) of an merged object , we can safely remove it (without loosing functionality).

b1rger:
> Second line meant that if we allow for overwriting the merge functionality in ontologies, the ontologies can also take care of storing whatever information they want to keep in whatever model they want ...

Oke, that should be #142

> ... to use by adding a pre-delete signal and checking for the issuer of that signal.

Not sure how signals are involved.

sennierer:
decided in Jfx:
- we will remove the labels
- however given that merges save information of merged objects in labels we keep it until the data can be moved somewhere else.

sennierer:
related to #242

PR #178, opened 2023-05-31
feat(apis_labels)!: drop apis_labels

b1rger:
This commit removes the apis_labels module:
* It deletes the module
* It removes the module from the INSTALLED_APPS list
* It modifies the detail_generic and edit_generic views in
  apis_entities, because they refer to the Label model
* It removes the label related code from the `merge_with` method in the
  TempEntityClass
* It removes label related codes from the generic serializers
* It removes a commented Label import from the metainfo model file
* It deletes apis_relations.forms - this only contained the
  EntityLabelForm for the Label model
* It removes the LabelTableBase and LabelTableEdit classes from
  apis_relations.tables
* It removes the apis_relations.forms module from the form_module_list
  in apis_relations.views, because the forms module does not exist
  anymore
* It removes the labels related stuff from the urls.py

Closes: #15


b1rger:
I'd like to move this forward, therefore adding the `needs-attention` label. Maybe everyone in @acdh-oeaw/prosnet could have a look if they use the labels and if so could think of a migration step.

b1rger:
We wait for this until we have a working approach to changes of models, probably using reversion

sennierer:
labels can be removed

PR #181, opened 2023-05-31
Feat: fixes #169 - edit and detail buttons in listviews

### LINKED ISSUE #169: Discuss `Use edit views` functionality
The user menu in the top navbar provides a `Use edit views` checkbox inside a form. It does not do anything and I didn't find a way of sending that form. Apparently it should set the session variable `edit_views`, which is checked in the list tables and then links the `name` column to the edit view of the entity instead of the detail view. Should we fix that or drop it or replace it with something else?

gregorpirgie:
**Describe your changes**
Add template columns with redirect to edit and detail views in entity listviews. 
Edit-view-column only shown if user is authenticated.

-> But anyway, listviews are currently only visible if user is logged in.
- so might be improved in the future by checking against actual edit permission, instead of only authentication. But this was beyond the scope of this issue for now. 

**Additional context**
Implemented as discussed in todays JFX.

- all other link columns are unchanged; I did only add the two new columns and changed the column order so that detail and edit columns are shown to the left. 

Resolves issue:
- #169 

### DIFF
--- a/apis_core/apis_entities/tables.py
+++ b/apis_core/apis_entities/tables.py
@@ -26,7 +26,7 @@ class MergeColumn(tables.Column):
         return mark_safe(input_form.format(value, value))
 
 
-def get_entities_table(entity, edit_v, default_cols):
+def get_entities_table(entity, edit_v, default_cols, is_authenticated=False):
     if default_cols is None:
         default_cols = [
             "name",
@@ -34,6 +34,7 @@ def get_entities_table(entity, edit_v, default_cols):
 
     class GenericEntitiesTable(tables.Table):
         def render_name(self, record, value):
+         
             if value == "":
                 return "(No name provided)"
             else:
@@ -71,7 +72,7 @@ def get_entities_table(entity, edit_v, default_cols):
 
         class Meta:
             model = caching.get_ontology_class_of_name(entity)
-            fields = ["detail"]+[default_cols]
+            fields = ["detail", "edit"] + default_cols if is_authenticated else ["detail"] + default_cols
             attrs = {"class": "table table-hover table-striped table-condensed"}
             # quick ensurance if column is indeed a field of this entity
             for col in default_cols:
@@ -85,8 +86,15 @@ def get_entities_table(entity, edit_v, default_cols):
         def __init__(self, *args, **kwargs):
 
             self.base_columns["detail"] = tables.TemplateColumn(
-                template_name='apis_entities/tables_detail_button_tempalte.html'
+                template_name='apis_entities/tables_detail_button_template.html',
+                extra_context={'entity_class':entity}
             )
+
+            if is_authenticated:
+                self.base_columns["edit"] = tables.TemplateColumn(
+                    template_name='apis_entities/tables_edit_button_template.html',
+                    extra_context={'entity_class':entity}
+                )
 
             super().__init__(*args, **kwargs)
 
--- a/apis_core/apis_entities/tables.py
+++ b/apis_core/apis_entities/tables.py
@@ -71,7 +71,7 @@ def get_entities_table(entity, edit_v, default_cols):
 
         class Meta:
             model = caching.get_ontology_class_of_name(entity)
-            fields = default_cols
+            fields = ["detail"]+[default_cols]
             attrs = {"class": "table table-hover table-striped table-condensed"}
             # quick ensurance if column is indeed a field of this entity
             for col in default_cols:
@@ -83,6 +83,11 @@ def get_entities_table(entity, edit_v, default_cols):
                     )
 
         def __init__(self, *args, **kwargs):
+
+            self.base_columns["detail"] = tables.TemplateColumn(
+                template_name='apis_entities/tables_detail_button_tempalte.html'
+            )
+
             super().__init__(*args, **kwargs)
 
     return GenericEntitiesTable
--- a/apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
+++ b/apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
@@ -1,0 +1,2 @@
+<!-- <button class="btn btn-primary btn-sm" onclick="updateRelationDate(this, '{{record.pk}}')">DetailView</button> -->
+<a class="btn btn-primary btn-sm" href="{{% url 'apis:generic_entities_detail_view' pk='{{record.pk}}' %}}" >Edit Views</a>
--- a/apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
+++ b/apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
@@ -1,2 +1,3 @@
-<!-- <button class="btn btn-primary btn-sm" onclick="updateRelationDate(this, '{{record.pk}}')">DetailView</button> -->
-<a class="btn btn-primary btn-sm" href="{{% url 'apis:generic_entities_detail_view' pk='{{record.pk}}' %}}" >Edit Views</a>
+<a
+    href="{% url 'apis:apis_entities:generic_entities_detail_view' entity=entity_class pk=record.pk %}"><i
+        data-feather="eye" title="eye"></i></a>
--- a/apis_core/apis_entities/templates/apis_entities/tables_edit_button_template.html
+++ b/apis_core/apis_entities/templates/apis_entities/tables_edit_button_template.html
@@ -1,0 +1,1 @@
+<a class="btn btn-primary btn-sm" href="{{% url 'apis:generic_entities_edit_view' pk='{{record.pk}}' %}}" >Edit Views</a>
--- a/apis_core/apis_entities/templates/apis_entities/tables_edit_button_template.html
+++ b/apis_core/apis_entities/templates/apis_entities/tables_edit_button_template.html
@@ -1,1 +1,9 @@
-<a class="btn btn-primary btn-sm" href="{{% url 'apis:generic_entities_edit_view' pk='{{record.pk}}' %}}" >Edit Views</a>
+<a
+    href="{% url 'apis:apis_entities:generic_entities_edit_view' entity=entity_class pk=record.pk %}">
+    <svg width="24" height="24" viewBox="0 0 24 24"
+        xmlns="http://www.w3.org/2000/svg" fill="none" stroke="currentColor"
+        stroke-width="2" stroke-linecap="round" stroke-linejoin="round"
+        class="feather feather-edit">
+        <path d="M11 4H4a2 2 0 0 0-2 2v14a2 2 0 0 0 2 2h14a2 2 0 0 0 2-2v-7" />
+        <path d="M18.5 2.5a2.121 2.121 0 0 1 3 3L12 15l-4 1 1-4 9.5-9.5z" />
+    </svg></a>
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -114,7 +114,6 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
         settings, "APIS_LIST_VIEW_TEMPLATE", "apis:apis_entities/generic_list.html"
     )
     login_url = "/accounts/login/"
-    print("class GenericListView New called")
     def __init__(self, *args, **kwargs):
         super(GenericListViewNew, self).__init__(*args, **kwargs)
         self.entity = None
@@ -171,6 +170,7 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
         else:
             edit_v = False
 
+        user_is_authenticated = self.request.user.is_authenticated
         selected_cols = self.request.GET.getlist(
             "columns"
         )  # populates "Select additional columns" dropdown
@@ -181,7 +181,7 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
         default_cols = default_cols + selected_cols
 
         self.table_class = get_entities_table(
-            class_name, edit_v, default_cols=default_cols
+            class_name, edit_v, default_cols=default_cols, is_authenticated=user_is_authenticated
         )
         table = super(GenericListViewNew, self).get_table()
         RequestConfig(
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -114,7 +114,7 @@ class GenericListViewNew(UserPassesTestMixin, ExportMixin, SingleTableView):
         settings, "APIS_LIST_VIEW_TEMPLATE", "apis:apis_entities/generic_list.html"
     )
     login_url = "/accounts/login/"
-
+    print("class GenericListView New called")
     def __init__(self, *args, **kwargs):
         super(GenericListViewNew, self).__init__(*args, **kwargs)
         self.entity = None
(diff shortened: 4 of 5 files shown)


koeaw (commented):
Without having looked at the code in detail yet, a few things which immediately stuck out:

* If the two columns were given a fixed width, preferably the same width, it'd make for a better user experience because the icons wouldn't change position from list view to list view.
* Users don't think in terms of Django views, they think in terms of "what does this do" (what is this for, which action does this icon represent,...). So, "View" would make for a better name for the first column/action than "Detail", _at least_ anywhere the name is surfaced to users, like the column header and anywhere in the resulting HTML.
Neat side-effect: matching word length will make the columns more visually pleasing. (Based on character count alone, if not actual width.) This'll also help with assigning the columns the same width.
* The two columns shouldn't be sortable. Right now, the column headers are clickable links ending in `/list/?sort=detail` and `/list/?sort=edit`, respectively.

---

Examples which demonstrate the icons/columns "jumping around" on the page from one entity to the next due to different column counts:

Table with many default columns looks "ok":
Screenshot 2023-05-31 at 22-06-44 Browse Werk

Table with few default columns has redundant whitespace/has the new columns oddly spaced apart :
Screenshot 2023-05-31 at 22-07-14 Browse Thema

---

That said, I think this update to the table is great!

apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
koeaw, line 1: I don't know why the two templates are implemented differently – one uses an `<i>` element for the icon, the other embeds the `<svg>` directly – but I'd make them follow the same logic.

The tag hierarchy should be indicated with indents throughout, i.e. closing tags should mirror opening tags for readability. Also, the templates seem to be missing newlines at the end of the file.

apis_core/apis_entities/tables.py
```diff
@@ -83,6 +84,18 @@ class Meta:
...
+                template_name='apis_entities/tables_detail_button_template.html',
+                extra_context={'entity_class':entity}
+            )
+
+            if is_authenticated:
+                self.base_columns["edit"] = tables.TemplateColumn(
+                    template_name='apis_entities/tables_edit_button_template.html',
+                    extra_context={'entity_class':entity}
+                )
+
```
koeaw, line 98: Is there an advantage to declaring the columns in the `__init__` method rather than in the class itself?

I'll add my main point of reference is [what I did](https://github.com/acdh-oeaw/apis-core-rdf/blob/dev/apis_core/apis_relations/tables.py#L59) in `PropertyTable` after looking at other table of ours + studying the docs, which didn't include many examples of `__init__` in use.

apis_core/apis_entities/tables.py
```diff
@@ -71,7 +72,7 @@ def render_name(self, record, value):
 
         class Meta:
             model = caching.get_ontology_class_of_name(entity)
-            fields = default_cols
+            fields = ["detail", "edit"] + default_cols if is_authenticated else ["detail"] + default_cols
```
koeaw, line 75: A cleaner way of checking for the right permissions – without inline `if`s (which I personally am not a fan of because they make the code harder to read) and without redundancy – might be to add:

```
if not is_authenticated:
    exclude = ["edit"]
```

to the `Meta` class.

b1rger (changes requested):
Please squash the commits
Please title the resulting commit(s) according to conventional commits (i.e. `feat(apis_entities): description...`)
Please fix the linter errors ;)

apis_core/apis_entities/tables.py
gregorpirgie: This is a good point, I considered that but your comment made me re-think why I did that overall. I agree with that it could (and should!) be more readable, but actually, I think that using fields here is semantically flawed from the beginning and was a mistake from my side anyway. 

Because: 

- fields are fields that come from the model - but the template columns are not part of the model(s)
    - side note: therefore I would protect against possible clashes with model-fields that might be named edit or detail by renaming the columns "_edit" and "_detail" and setting a verbose_name on the columns (would use EditView and DetailView as you suggested @koeaw)
- I used fields here, for the sole purpose to force the order of the columns  (by inserting the column names before  the existing ones). 
- a better way (more semantically clear way) to achieve that ordering would be to use the `sequence = ("detail", "edit", "...")` attribute in the Meta-Class. - see [tables2 docs](https://django-tables2.readthedocs.io/en/latest/pages/swapping-columns.html)

So I would suggest to be very expressive by setting the meta class like this: 

```python 

class Meta:
    model = caching.get_ontology_class_of_name(entity)
    fields =  default_cols 
    if is_authenticated:
        sequence = ('_detail', '_edit', '...') # note: the '...' is special syntax here to add all other columns (fields) in the order they where defined. See the table2 docs mentioned above (same link)
    else: 
       sequence = ('_detail', '...')
```

I think we can't get rid of all redundancy, because: as "_edit" and "_detail" are not fields present in the models, they need to be added to the `fields` attribute or the `sequence` attribute anyway, before I can use `exclude`, as you suggested. So when using `fields` I would need to always add both (for example in default_cols) and then remove "_edit" conditionally, as you suggested. Otherwise, Django throws an error if the template column does not exist or a field with that name.

apis_core/apis_entities/templates/apis_entities/tables_detail_button_template.html
gregorpirgie: Will change that, forget about it after I copy pasted how the icons are implemented at other places in the code for the first draft!

apis_core/apis_entities/tables.py
gregorpirgie: I just applied the same conventions that I found in other uses of this implementation in the existing code base (and how it was explained in the docs). 

Again, you made me think about it. I would leave it as it, but actually I am not 100% sure how this part of the code runs. Here are my assumptions, based on my current knowledge:

- the class is nested in a factory function that takes over the creation of the class itself. 
- this matters, I think, because the class itself is constructed by  a special metaclass. What you define within the inner `class Meta` is stuff that will be run when the actual class is constructed (not the instances of the class, which happens in the init).
- usually, metaclasses construct their classes when a python program starts up, and only ONCE. So actually, at this point, no user or request would be present and it would not make sense (I think!) to include the check against the permissions within the `class meta`block. 
- This is bypassed by the fact that the whole class is wrapped in a factory method. This delays the actual construction of the class and also the init of the classes instance(s) to the moment the factory method is called, and so this construction is run each time an instance of the class is requested (which implies that for each request of an instance, the whole class is reconstructed).
- at least, that's my guess. And that would explain why the permission check in the  `class  meta` body actually works in he current implementation.

So I will test my assumptions now. because, if they are correct, this is whole pattern could be improved. As this is sort of defeating the purpose of generic classes and OOP in general. We are creating a new class and a new instance of that class now each time we want to render a single table. As this is not the only place in the codebase where this factory function + metaclass construction pattern occurs, improving on this pattern might be worth it. And it would help with understanding how the code is actually run.

Based on those assumption, to answer your question: 
- I guess in the current pattern it would not matter if the template columns are added in the init or the class body. because, as I said, construction of the class and initialising an instance happen sequentially in this code each time.
- but adding the templates to the init seems still to be more correct to me, as this is stuff should actually be bound to the instance of the table, not the whole class. 
- also, if this pattern might be refactored in the future, having the templates in the class body might break the code or create idiosyncratic behaviour.

Again, this is written from a couple of based assumption that I have about the inner workings of this particular code. Will check them now and update you.

gregorpirgie: I just tested it, my assumptions seem to be correct. 

I copied the code of the table class and pasted it outside the factory function. I then bound this table to a certain entity-model class and added a condition in the view serving the listviews if this entity is requested, to use my TestTable. 
I added print statements in the `__init__` and `class meta` bodies to both implementations and requested the list view of this specific entity and a generic other one in the frontend. 

Here is what django prints: 

```bash
Performing system checks...

meta reached in TestTable class # --> you can see that the TestTable outside the factory is constructed before anything else (this is the moment the class is constructed)

June 01, 2023 - 10:11:52
Django version 4.1.9, using settings 'apis_ontology.settings.server_settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.

# Here I request the listview of an entity using the factory function:
meta reached in table class # Class is constructed
init called in table # Instance is initialised
[01/Jun/2023 10:11:58] "GET /apis/entities/entity/e1_crm_entity/list/ HTTP/1.1" 200 13248
[01/Jun/2023 10:11:58] "GET /static/autocomplete_light/select2.js HTTP/1.1" 200 4935
Not Found: /apis/entities/entity/e1_crm_entity/list/piwik.js
[01/Jun/2023 10:11:58] "GET /apis/entities/entity/e1_crm_entity/list/piwik.js HTTP/1.1" 404 1599

# Here I request the entity that uses the table class outside the factory function:
init called in TestTable # --> only an instance is created at runtime, the class is re-used from before startup! 

[01/Jun/2023 10:11:59] "GET /apis/entities/entity/e55_type/list/ HTTP/1.1" 200 7939
[01/Jun/2023 10:11:59] "GET /static/autocomplete_light/select2.js HTTP/1.1" 304 0
Not Found: /apis/entities/entity/e55_type/list/piwik.js
[01/Jun/2023 10:11:59] "GET /apis/entities/entity/e55_type/list/piwik.js HTTP/1.1" 404 1597
```

I think it does no make sense to create a class and an instance of a class each time you want to use a generic class. This is a separate issue, but I already have ideas how to resolve that!

gregorpirgie: @b1rger  @sennierer @koeaw  - This specific thread is important, please have a look. 

Thanks for making me think about it again, @koeaw !

gregorpirgie: So, kay, to answer your question once more: that's why I think placing the template columns in the __init__ block is the correct approach, even if we use the factory function. because the conditions should be applied at runtime each time an instance of the table (listview) is requested, on a per user basis.

gregorpirgie:
@b1rger will do all of that!

gregorpirgie:
@koeaw regarding the layout issue: 

I would not fix that in this PR, as it has nothing to do with the new columns, but applies to all columns (so this is a separate issue). The same thing happens if you have two small columns in a table, etc. It just became visible with this specific implementation, but existed before.

apis_core/apis_entities/tables.py
gregorpirgie: Created a new issue for re-thinking / refactoring the factory pattern used across the codebase, see: #182 #182

gregorpirgie: Created a new issue for re-thinking / refactoring the factory pattern used across the codebase, see: #182

apis_core/apis_entities/tables.py
koeaw: Just a quick note: 

> (would use EditView and DetailView as you suggested @koeaw)

What I actually meant was that the _words shown to users_ should read "Edit" (-> edit this data item) and "View" (-> view the full item referred to by this row). But to clarify internally what the columns point to, i.e. (which) Django views, sounds like a good idea as well.

"Show" was the only other word I could think of for the eye column, but I think "View" fits better semantically. + It should be possible to prevent potential confusion over "View" as verbose name vs. "View" for Django views with the help of code comments for devs.

gregorpirgie:
> I wouldn't change anything about the layout of the other columns. It's fine that they are of variable width because we can't know beforehand how much space a given field's contents will or should take.
> 
> These two new columns are the only ones for which we can provide a fixed width (based on col header size) because their contents won't change.

Will try to implement that, you are right!

gregorpirgie:
General Update: 

I have now looked deeper into how django_tables2 works (the library we use to render all tables). I came to the conclusion that we have used the library not how it is intended to be used (I am/was also guilty of that...). 

It provides are clear distinction between what should and can only be defined on a table class (within the `class meta`block) and what can and should be defined on an per-instance basis. We totally mixed this up and furthermore, created a tight coupling between our views and the tables (which is even worse, I would say).

I will work this understanding now into the current implementation of this PR, specifically for the tables in apis_entities. 

And this time, instead of re-inventing the wheel, I will make use of the already implemented API in django_tables2. 

This will include a refactor of the factory function for entities-tables, a refactor of the listviews that generates the table, and everything that was already mentioned in this PR. Implementing the columns that where the initial scope of this PR properly depends on this refactor, that's why I will include it now. 

If there are no objections. But I will close this PR for now, and create a new branch with a new PR that references the initial issue as part of the implementation, once my work is finished.

---> Feel free to re-open this PR if you disagree! <----

apis_core/apis_entities/tables.py
koeaw: There is no advantage to using `EditView` over `Edit`, only disadvantages from a UI/UX POV.

Interfaces should be well structured, easy to understand, easy to navigate.

Individual elements/parts shouldn't be distracting, take up unnecessary space or interfere with a user's flow in other ways. These two columns are navigational aids that use up space that was previously reserved for the thing that users are actually concerned with (content), so they shouldn't draw even more attention. There is no way around forcing the columns to be wider than their contents of _1 icon_ because we can't have _no_ header, but to broaden them artificially by making the headers even longer (even worse, by using formatting users don't encounter IRL – camel case is not a thing in natural language) when short words are perfectly fine or even self-explanatory does not help users. The suggestion to give the columns a fixed width, and make them _the same width_ (including the word in the header), is part of that, too.

b1rger: I think using the terminology `EditView` or `DetailView` in the user interface is just confusing. If you take github as an example, they simply use `Edit` for the comments we are writing and not `EditView` or `EditForm`. I think the table columns describe *actions* and actions are verbs. So IMO it should be `Edit` and `View`.

But I also think that the table header could be left empty - we are using the same icons in other places without a description what they mean. I would be happy without a table header for those two columns, but there should definitely be a tooltip when hovering over the icons.

b1rger:
> This will include a refactor of the factory function for entities-tables, a refactor of the listviews that generates the table, and everything that was already mentioned in this PR. Implementing the columns that where the initial scope of this PR properly depends on this refactor, that's why I will include it now.

I just want to point out that you went from *adding two columns in a table* to *refactor everything*. I think both are valid issues and should be handled separately, especially given how much discussion there is already regarding those two columns.
But there is also the fact that the columns are a low priority issue, so I'm not going to block the integration in a separate bigger PR. But be aware that even PRs for one line changes take more than 6 weeks to be reviewed 🥲

apis_core/apis_entities/tables.py
gregorpirgie: Alright, I would draft it now without headers, as this should solve both issues (spacing and naming). 

And in a tooltip we can be more precise what we mean ('edit entity' and 'view entity detail') or whatever.

gregorpirgie:
Obviously true, but I really think it is necessary and part of the implementation of the columns itself. There is so much wrong with this little snippet, that just adding two columns to it and leaving the rest untouched just adds to the mess the codebase already is (in some parts...). 

And I am not going to refactor everything, just the tables in apis_entities (and because they are tightly coupled, everything they are coupled to, but only those parts of the listview (for example) that relate to the tables.  Because we do stuff in tables.py that should happen in the view and we do stuff in the view that we should do in the tables.

This is also an investment in the future, as the django_tables2 library provides an flexible api, that our faulty implementation broke (in the sense that we can no longer use it's flexibility, because we hardcoded unnecessary stuff in the wrong place). So adding features in the future will be way easier, once it is setup correctly. And I am almost done already.

Adding new columns to the existing tables is something that I have done multiple times across projects in the last two years, so yeah, the two columns are a low priority, but the tables themselves should not be. It will come up again, I am pretty sure.

Issue #182, opened 2023-06-01
Possible Refactor: re-think factory methods that create classes and class instances


gregorpirgie:
As an introduction, please read the discussion in PR #181 under the question about the __init__ function used in the apis_entities.tables.py factory function: [link to thread](https://github.com/acdh-oeaw/apis-core-rdf/pull/181#discussion_r1212349441).

# Description of the current pattern: 

We use factory methods at several places in our code-base when generic versions of specific classes are needed.  (see a list of the places where it is used below).
But in the current implementation, the whole class and a single instance are created each time we would actually just need an instance of the generic class. 

Also, there is often no clear distinction between code that should be used on a per-instance basis and computed at runtime (request time of an instance) and code that is applied to the whole class.

# What is the issue with that: 

- the code is harder to reason about, because we treat classes like instances and we can do that because the class is re-created on each request and then garbage-collected
- specifically, we have instance-scoped logic in the class body, that should actually live in the init method.
- but also we re-create classes and instances for each request (after which they are garbage collected!), which is neither resource-aware, nor does it follow the OOP logic and the intended use of the libraries we make use of that provide those classes that we create at runtime.

# Suggested Refactor: 

- refactor the classes themselves and make clear distinctions between instance-scoped logic and settings and class-scoped logic and settings.
- either move the generic logic also in the init function, and get rid of the factory method (which in most cases should not be possible, I think, as generic settings are needed also in the `class meta block`)
- or leave the factory methods as they are (but still, refactor the class body to be more semantically and logically clear), but on startup, create all possible variants of the generic classes, store them in a dictionary or another appropriate datastructure and then each time an instance of a variant of the class is needed, retrieve the pre-created class from the storing datastructure (not the factory method) and create a new instance of it. 
- this way, the classes would stay alive and are not garbage collected. 

# Where the pattern is used atm :

These I could find: 

## apis_core/apis_relations
- in `rel_filters.py` - factory: `get_generic_relation_filter` ([see here](https://github.com/acdh-oeaw/apis-core-rdf/blob/c35d13bcf026b9946c3ec0eb908b87880a246e02/apis_core/apis_relations/rel_filters.py#L92))
- in `tables.py` - factory: `get_generic_relation_listview_table` ([see here](https://github.com/acdh-oeaw/apis-core-rdf/blob/c35d13bcf026b9946c3ec0eb908b87880a246e02/apis_core/apis_relations/tables.py#L135))

## apis_core/apis_metainfo
[...]

PR #185, opened 2023-06-05
Gregor/refactor/entity table factory

gregorpirgie:
**Describe your changes**

refactor: tables.py and entities-table-factory.

Overview:

Tables.py
- refactor of tables.py file to own python module within apis_entities
- refactor of get_entities_tables factory method into factory-class.
- factory-class persists already created table-classes, instead of
   re-creating them on each request.
- inner-generic-table-class removes constraints of former implementation
- its the broadest implementation of the table
- the calling view is now responsible for customizing the instance of
   the table class that will be send to the template, instead of
   hardcoding the customization in the class itself.
- added utility class get_entities_table,
that uses the new FactoryClass, but catches all
existing calls from other modules, to not break current implementations.

GenericListViewNew in apis_entities.views.py:
- table specific code moved into own class / mixin that is now
  inherited by the GenericListViewNew
- refactored variable names
- moved seperate logic into own methods
- added docstrings: with context information that is not visible
  from the code itself.
- tried to name methods in a way that they explain what they do
- added inline comments on class vars and instance vars that are
bound to implicit logic that happens in the parent classes - these
should NOT be renamed, or things will break.
- removed as many if else statement block-series or try-except 
  blocks that where even interdependent on one another as 
  possible. The prior version was, IMO, really hard to read 
  and reason about.

**Related issues and PRs**
- #182 - implements factory refactor as an example solution for apis_entities

Resolves issue:
- #169

---> sorry, my formatter doesn't work currently and the files are not formatted correctly. <--

- [ ] My changes follow the project's code formatting rules and style guidelines


gregorpirgie:
Styling the columns (adjusting the width) was not possible atm, this must be done in a separate PR. 

- headers are now removed for the two edit and detail columns. 
- edit and detail columns are not sortable and don't participate in sorting order
- obviously, the two columns are now also implemented, which resolves #169

b1rger (changes requested):
Not going into the content right now, I'd appreciate it if the structure of the commits would be a bit more cleaned up. If you want to change something you introduced in a prior commit, please don't simply create a new commit, but fix the commit itself. Adding additional commits with bug fixes makes reviewing harder and these changes not have to land into our main history. `Bug` is not an acceptable commit title. The `Style` commit should not be a separate commit - every commit should adhere to styling conventions. In fact, every commit should run cleanly run through all the automated tests.
So: please squash you commits. Please put yourself in the shoes of your reviewers.

b1rger:
Also, there are merge conflics. Please resolve them

gregorpirgie:
You are totally right, I didn't notice that there where unnecessary commits, that weren't meant to be published, in the branches history. Will rework that and remember it in future commits.

gregorpirgie:
Closing again, for new PR with squashed commits and formatted files.

PR #186, opened 2023-06-06
refactor: table implementation in apis_entities

### LINKED ISSUE #169: Discuss `Use edit views` functionality

gregorpirgie:
Third time's the charm.

Tables.py
- refactor of tables.py file to own python module within apis_entities
- refactor of get_entities_tables factory method into factory-class.
- factory-class persists already created table-classes, instead of re-creating them on each request.
- inner-generic-table-class removes constraints of former implementation
- its the broadest implementation of the table
- the calling view is now responsible for customizing the instance of the table class that will be send to the template, instead of hardcoding the customization in the class itself.
- added utility class get_entities_table, that uses the new FactoryClass, but catches all
existing calls from other modules, to not break current implementations.

GenericListViewNew in apis_entities.views.py:
- table specific code moved into own class / mixin that is now inherited by the GenericListViewNew
- refactored variable names
- moved seperate logic into own methods
- added docstrings: with context information that is not visible from the code itself.
- tried to name methods in a way that they explain what they do
- added hint in method docstring if method is inherited from superclass
- added inline comments on class vars and instance vars that are bound to implicit logic that happens in the parent classes - these should NOT be renamed, or things will break.
- removed as many if else statement block-series or try-except blocks that where even interdependent on one another as possible. The prior version was, IMO, really hard to read and reason about.

**Related issues and PRs**

Resolves issue:
- closes #169 

- [ ] My changes follow the project's code formatting rules and style guidelines


koeaw:
A general thing I wanted to make a note of: I'd suggest renaming `visible_columns` to something different, maybe `display_columns` or similar? Because "visible" implies a visible/hidden (or invisible) dichotomy, with the latter suggesting the stuff is effectively there but only hidden from view.

In HTML/CSS, too, visible/hidden means "visible to the eye"; hidden elements are still loaded/rendered/can be accessed.

gregorpirgie:
> A general thing I wanted to make a note of: I'd suggest renaming `visible_columns` to something different, maybe `display_columns` or similar? Because "visible" implies a visible/hidden (or invisible) dichotomy, with the latter suggesting the stuff is effectively there but only hidden from view.
> 
> In HTML/CSS, too, visible/hidden means "visible to the eye"; hidden elements are still loaded/rendered/can be accessed.
> 
> ETA: probably another case for a thing to put in sth. like a design doc/a (new) portion of the project docs dedicated to rules & conventions etc.

Yeah, good point. But this is a naming convention adopted from django_tables2, which differentiates between "visible" and "hidden" columns and each column has a "hide" attribute, that I actually have to trigger in the __init__ to "hide" the column.  The param visible_columns is custom still, but follows this naming convention for that reason. So, even if I rename the attribute to visible_columns, people might still be confused that this triggers a column.hide() for columns not in the list... So I would leave it as is.

b1rger (changes requested):
Please create your PR against `dev`, not `main`. Your commited `views.py` seems to contain a lot of unrelated changes, please clean that up (I usually have a look at the github diff before marking a PR as ready for review, to make sure the PR really only contains the changes I wanted to merge)

Issue #197, opened 2023-06-12
Show and use notes field in relations form


b1rger:
The TempTriple has a notes field. This notes field should be usable in the triple form

PR #198, opened 2023-06-12
fix(apis_relations): show notes field and store notes

b1rger:
The TempTriple model has a notes field, that should be usabel to store
metadata. This commit adds it to the form and the form handling of the
GenericTripleForm.

Closes: #197


gythaogg (approved):
Works neatly!

PR #216, opened 2023-06-21
Change ContentType.objects.get to .get_for_model

richardhadden:
**Describe your changes**

- Changed `ContentType.objects.get(model=<MODEL NAME>)` to `ContentType.objects.get_for_model(<MODEL>)`

**Additional context**
Due to case sensitivity issues in Postgres, looking up ContentTypes with the string name causes problems. Previously using the workaround script (https://github.com/acdh-oeaw/apis-ontologies/blob/8a561a50fee5220e0e93255c40cb152695531864/jelinek/ontology_specific_scripts/postgres_specific.py), but this causes further errors.

Instead, change the ContentType lookups to use the actual model, rather than the string name. There still seems to be some use of `ContentType.objects.get()` elsewhere; will investigate.

- [X] My changes don't generate new warnings or errors
- [X] My changes follow the project's code formatting rules and style guidelines

### DIFF
--- a/apis_core/apis_relations/models.py
+++ b/apis_core/apis_relations/models.py
@@ -122,7 +122,7 @@ def subj_or_obj_class_changed(sender, is_subj, **kwargs):
             class_current = contenttype_current.model_class()
             for class_child in class_current.__subclasses__():
                 # TODO: Avoid ContentType DB fetch
-                contenttype_child = ContentType.objects.get(model=class_child.__name__)
+                contenttype_child = ContentType.objects.get_for_model(class_child)
                 child_list.append(contenttype_child)
                 child_list.extend(get_all_children(contenttype_child))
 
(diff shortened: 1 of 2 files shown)


richardhadden:
Should check that this doesn't break anything with MariaDB instances (though I can't see why it would cause a problem)

b1rger (changes requested):
Please change your commit message title to adhere to the [conventionalcommits](https://www.conventionalcommits.org) standard.
Please put the description of the change (what you wrote in the pull request) into the commit message - if we ever leave github, we lose the information, but if it is part of the commit history, we keep it.
You can change the last commit in a branch using `git commit --amend`. You can push the updated commit using `git push -f`

richardhadden:
@b1rger Sorry! Will read the rules more thoroughly!

apis_core/apis_relations/models.py
```diff
@@ -292,7 +292,7 @@ def get_all_childs(cls_current):
         if self.subj is not None:
             subj_class_name = self.subj.__class__.__name__
             if (
-                ContentType.objects.get(model=subj_class_name)
+                ContentType.objects.get_for_model(self.subj.__class__)
                 not in self.prop.subj_class.all()
             ):
```
sennierer, line 297: test

richardhadden:
Is this ok to merge now? I can't use Postgres without this fix

PR #231, opened 2023-06-29
refactor(apis_metainfo)!: drop `Source` model

b1rger:
The `Source` model was introduced to store information about where an
entity was imported from. This can easily be replaced by a custom field
in the ontology. Therefore the `Source` model was removed.

Closes: https://github.com/acdh-oeaw/apis-core-rdf/issues/227


koeaw:
Since we're using `Source` for Frischmuth and I won't have time to work on a replacement for it this week, I would ask that merging this be postponed until next week.

sennierer:
given that @koeaw asked for postponing the merge convertged this to a draft for the moment

b1rger:
2 months have passed. We do now have releases, so merging this would not interrupt downstream projects if they don't bump versions. Can we move this forward?

CC: @acdh-oeaw/prosnet

PR #232, opened 2023-06-29
chore(black): ignore migrations

b1rger:


pyproject.toml
```diff
@@ -65,3 +65,11 @@ extension="html"
 blank_line_before_tag="block,if"
```
sennierer, line 65: regarding #236, can we not just remove the if here, so that no blank line is added before if statements?

sennierer (commented):
Is that by intention just a draft? I think black shouldnt tinker with auo-generated migration files.

pyproject.toml
b1rger: Sure, but it took us months to get to this basic set of `djlint` settings

b1rger:
> Is that by intention just a draft? I think black shouldnt tinker with auo-generated migration files.

Yes, thats on purpose. The existing migrations all adhere to `black` and I wasn't sure if simply running black on new migrations isn't the easier fix. If we ignore migrations, we ignore them also on the ones we maybe write ourselves at some point

b1rger:
I'm closing this. Thinking about it, it makes sense to have our whole codebase follow the same rules - migrations are autogenerated at the moment, but they might be written manually at some point. Running `black` on them does not hurt, IMO

PR #240, opened 2023-07-03
refactor: delete unused files and code

b1rger:
The files in apis_entities/
* `autocomplete_light_registry.py`
* `custom_authentication.py`
* `custom_autocompletes.py`
are not being used anywhere

The files in apis_relations/
* `autocomplete_light_registry.py`
* `serializers.py`
are not being used anywhere

The file `apis_relations/forms.py` was only used in
`apis_relations/views.py` and in there the result of the function that
used it was assigned to a variable that was not used anywhere. The whole
code block is now deleted.


apis_core/apis_entities/custom_authentication.py
sennierer, line 1: I think we should keep that one and add it as the default authentication class to the restframework setting. That allows to use the token also in the query params.

apis_core/apis_relations/autocomplete_light_registry.py
```diff
@@ -1,257 +0,0 @@
...
-
-path_ac_settings = getattr(settings, "APIS_AUTOCOMPLETE_SETTINGS", False)
-if path_ac_settings:
-    ac_settings = importlib.import_module(path_ac_settings)
-    ac_settings = getattr(ac_settings, "autocomp_settings")
-else:
-    from apis_core.default_settings.NER_settings import autocomp_settings as ac_settings
-
-
-class StanbolAutocompleteBase(al.AutocompleteListTemplate):
```
sennierer, line 26: That function provided the autocomplete suggestions for entities derived from our stanbol instance. Given that we dont have a fixed set of entity types anymore this does not make much sense anymore. However, we want to indlude a way in apis-rdf-core to hook additional autocomplete sources to the entities autocompletes. E.g. a list of tuples in the settings with the entity and the function that recieves the query string and returns a configurable amount of suggestions in a defined format.

sennierer (approved):
there are some things (see comments) that we will need later on. However, for the moment its fine to remove them.

apis_core/apis_entities/custom_authentication.py
b1rger: Moved it to `apis_core/utils/authentication.py`

Issue #263, opened 2023-08-29
filters in generic API are not working anymore


sennierer:
filters in the generic API routes are not applied, all objects are returned
In Tibschool instance is working as intended, using this commit (https://github.com/acdh-oeaw/apis-core-rdf/commit/dff296b)

sennierer:
@stefanprobst thats the issue tracking the broken filters

PR #264, opened 2023-08-29
fix: use the filter method of the superclass, if it exists

b1rger:
Python does not pass the object from one parent classes' method to the
next, we have to call super() ourselves, if the method exists.

Closes: #263


sennierer (approved):
works now as expected
Thanks for the fix!

PR #276, opened 2023-09-08
fix(apis_entities, apis_metainfo)!: drop texts

b1rger:
Closes: #228, #7


sennierer:
we remove the text and create a issue in the highlighter repo about broken functionality ( @sennierer )

PR #284, opened 2023-09-13
feat: new relation implementation

b1rger:
# Installation
Add the app to Djangos INSTALLED_APPS: `INSTALLED_APPS += ["apis_core.relations"]`

Define relations like this:

```Python
class PersonHasLivingPlace(Relation):
    subj_model = Person
    obj_model = Place
    name = "bewohnt"
    name_reverse = "Bewohner von"
```

The app uses `htmx` - but it should also work without it. Please test with and without.

## Templatetags

There are a couple of templatetags to make the live with Relations easier:

`relations_table` gives you a list of relations. You can use the argument `instance` to limit to relations a specific instance is part of. You can use the argument `tocontenttype` to limit the list to relations that a have a specific contenttype as object.

`instance_is_related_to instance` returns a list of Contenttypes this instance *is* related to.

`relations_links` creates a list of links to relations views. The relations view lists relations and provides a form to create new ones. You can use the argument `instance` to limit to relations a specific instance is part of. You can use the argument `tocontenttype` to limit to relations that have a specific contenttype as object.

`instance_can_be_related_to` lists the contenttypes of all models an instance can be related to. You have to pass the `instance` as an argument.

### Usage

In the existing APIS templates, you can use this block to override the relations in `detail_generic.html`:
```
{% block relations %}
{% instance_is_related_to instance=object as contenttypes %}
{% for contenttype in contenttypes %}
<div class="card card-default">
  <div class="card-header" role="tab" id="heading{{ forloop.counter }}">
    <h4 class="card-title">
      <a role="button">{{ contenttype.name }}</a>
    </h4>
  </div>
    <div role="tabcard">
       <div class="card-body">
        {% relations_table instance=object tocontenttype=contenttype as table %}
	{% render_table table %}
    </div>
    </div>
  </div>
{% endfor %}
{% endblock %}
```

In the existing APIS templates, you can use this block to override the relations in `edit_generic.html`:
```
{% block relations %}
{% instance_can_be_related_to instance as contenttypes %}
{% for contenttype in contenttypes %}
  <div class="card card-default">
    <div class="card-header" role="tab" id="heading{{ forloop.counter }}">
      <h4 class="card-title">
        <a role="button"
           data-toggle="collapse"
           data-parent="#accordion"
           href="#collapse{{ forloop.counter }}"
           aria-expanded="true"
           aria-controls="collapse{{ forloop.counter }}">{{ contenttype.name }}</a>
      </h4>
    </div>
    <div id="collapse{{ forloop.counter }}" class="card-collapse collapse" role="tabcard" aria-labelledby="heading{{ forloop.counter }}">
      <div id="tab_{{ obj.2 }}" class="card-body">
      {% relations_table instance=instance tocontenttype=contenttype as table %}
      {% render_table table %}

      {% relations_links instance=instance tocontenttype=contenttype htmx=True %}
    </div>
  </div>
{% endfor %}
{% endblock relations %}
```

# Migration from TempTripe

There is also a migration path from `TempTriple`: a `post_save` signal for `TempTriple` `get_or_create`s a `Relation` with the same types and the same `name` and `name_reverse` as the `TempTriple`. It then copies the field data from the TempTriple to the Relation. This also works on updates of the TempTriples.

Example:
```
class PersonHasLivingPlace(Relation, LegacyDateMixin):
    subj_model = Person
    obj_model = Place
    name = "bewohnt"
    name_reverse = "Bewohner von"

    # Those are only for the Migration from TempTriple to Relation
    temptriple_name = "bewohnt"
    temptriple_name_reverse = "Bewohner von"
    temptriple_field_list = ['start_date', 'start_start_date', 'start_end_date', 'end_date', 'end_start_date', 'end_end_date', 'start_date_written', 'end_date_written']
```


b1rger:
Discussion points:
- [ ] Should the `Relation` model inherit from `RootObject`
- [ ] Should the obj and subj fields be part of the Relation itself or only of the childmodels (the former lets us list all existing relations using one sql table; the latter lets us have one less tabel - instead of `rootobject` + `relation` + `specificrelation` we only have `rootobject` + `specificrelation` which also mimics what we do with entities

sennierer:
Working on the fix of the reversions (#161) and the possibility to use the reversions for the management of versions of entities for citing (with new versions on all changes) there is one more thing to consider. Having the FKs in the relation itself would allow to use manytomany fields on the Person (or any other) entity and `follow` that relations with the versions tool...

gythaogg:
@b1rger 
I ran into `django.core.exceptions.FieldError: Cannot resolve keyword 'metadata' into field.`
This is coming from the post save signal for temptriples - but the metadata field is missing in Relations base model
It seems like a JSONField 
`rel, created = model.objects.get_or_create(subj=instance.subj, obj=instance.obj, metadata={"temptriple": instance.pk})`

gythaogg:
The Object field while adding/editing a relationship was an autocomplete field - in this PR it's become drop down, which is not as convenient when there are hundreds of entities.

gythaogg:
### Bugs

1. htmx delete does not work (CORS violation error, maybe the CSRF token is not passed in the header in this request?) 
Confirmation dialog appears and clicking OK does nothing. However pasting the delete URL in another window shows the same dialog in the body with APIS/Relations styles and clicking OK there does delete the relation.
2. Edit relation link does not care about htmx - always opens in a new window; ideally it should expand inside an element within the relationship row and not reload the page on save.
3. When entities are merged, its relations should also be merged accordingly 
4. The endpoint `/apis/relations/all` throws an  `Attribute error - type object 'Relation' has no attribute 'name'`

### Open points and feature requests

1. What should happen to a relation when an entity in its subject or object is deleted? At the moment the deleted entity will appear as NULL in the relation.
2. Could we expect a paginated view of relations or should it be left to the ontologies to customise in their templates?
3. The methods `obj_to_subj_text` and `subj_to_obj_text` expect `name` and `reverse_name` in order to work, it checks for the presence of these fields but when they are not present they return a generic "relation to" as the name of the relationship. Could we enforce these fields from the Relations base class?
4. While listing relationships in the entity detail view, could we always list the entity in question as the subject and use the reverse_name when the entity appears as an object in a relationship?
5. Previously the "other entity" (always displayed as an object in the detail page) was a link and this is a really useful feature. In order to do this, it would be nice to have more customisable list views (at the moment I am going to try to override template tags and see if I can achieve it for Tibschol)
6. Post save signal that copies triples into the new model - may be unnecessary if the migration is done using a management command from triples to the new relations. I did not try this feature as in my edit and detail view templates I replaced the existing relationship block with a new block that makes use of this PR. So I don't have access to the triples anymore in the detail view...

@b1rger Thank you very much for the PR - I know this is a long list, but I really appreciate the new implementation and will be happy to work with you to fix them.

gythaogg:
@b1rger Could we include the relation_css block in the list template so that if I override the relations.css file in my ontology the styles are picked up? At the moment I've overridden the list template to include the relations_css templatetag so that my styles are picked up

gythaogg:
@b1rger in relations/models.py the assumed field names for the forward relation and the inverse are **name** and **reverse_name**
```
@property
def obj_to_subj_text(self):
  if hasattr(self, "reverse_name"):
    return f"{self.subj} {self.reverse_name} {self.obj}"
  return f"{self.obj} relation to {self.subj}"
```
however in the relation_links template for listing the allowed relationships between the entity and another entity of the same type the code expects `name_reverse`
```     
{% if rel in relation_types_reverse %}
        {{ model.name_reverse }}
      {% else %}
        {{ model.name }}
{% endif %}
```

b1rger:
I've rebased on current main and fixed the merge conflicts. I'm not sure which of the mentioned bugs are still valid, @gythaogg maybe you can strike through the fixed bugs. It might also be a discussion point if we want to limit our relations to children of `RootObject`

b1rger:
superseded by https://github.com/acdh-oeaw/apis-core-rdf/pull/1062

Issue #221, opened 2023-06-22
URIs - one per entity or multiple?


gythaogg:
When adding a relation, load_subj_obj_prop seems to assume that an object can have only one Uri object - Line 174 `Uri.objects.get(root_object=entity_instance_other)`, but I imagine the intention is to allow multiple URIs for an entity (as far as I can infer from the detail view that has the label "URI(s")

sennierer:
https://github.com/acdh-oeaw/apis-core-rdf/blob/dev/apis_core/apis_relations/forms2.py#L147 needs to be changed to `Uri.objects.get(root_object=entity_instance_other, uri__startswith=settings.get("APIS_BASE_URI"))`

b1rger:
We hit that bug too and it should be:
```
Uri.objects.filter(root_object=entity_instance_other, uri__startswith=getattr(settings, "APIS_BASE_URI", "")).first()
```

b1rger:
It would be great if that would be fixed

PR #288, opened 2023-09-20
Consider only the first primary URI of entities

gythaogg:
where an entity might have more than one URI

primary URI is one that is in the same domain as
defined in APIS_BASE_URI in settings

the default value for APIS_BASE_URI is hardcoded to be consistent with the same value used in apis_entities at the time of the creation of the URI object

Resolves issue:
- #221 

- [x ] My changes don't generate new warnings or errors
- [x ] My changes follow the project's code formatting rules and style guidelines

### DIFF
--- a/apis_core/apis_relations/forms2.py
+++ b/apis_core/apis_relations/forms2.py
@@ -145,7 +145,14 @@ class GenericTripleForm(forms.ModelForm):
         self.fields["property"].choices = [property_initial_value]
 
         other_entity_initial_value = (
-            str(Uri.objects.get(root_object=entity_instance_other)),
+            str(
+                Uri.objects.filter(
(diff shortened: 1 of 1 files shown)


b1rger (changes requested):
Please format the commit message title according to the [conventional commit standard](https://www.conventionalcommits.org/en/v1.0.0/). I.e. `fix(apis_relations): fix entity lookup for multi URI entities`

Please add `Closes: #221` to the commit message body.

There are a couple of newlines removed from the file, those changes have nothing to do with the fix itself. Ideally those would be removed (Probably some setting from your editor?)

Other than that - LGTM!

b1rger (approved):

b1rger:
Thanks!

Issue #222, opened 2023-06-22
Move `apis` module to a separate repository


b1rger:
The `apis` module is bascially only a collection of custom acdh settings and a base `urls.py` file. I suggest to move the module to a separate repository `apis-acdh-defaults`. The various instances would then simply have to define a dependency on the module and replace the `from apis.settings.base import *` with a `from apis_acdh_defaults.settings.base import *`

PR #300, opened 2023-09-25
refactor!: drop apis module

b1rger:
The `apis` module provided default settings for apis instances at the
ACDH-CH. It was moved to a separate repository and lives now under
https://github.com/acdh-oeaw/apis-acdhch-default-settings

This commit therefore removes the `apis` module. The `apis` module also
contained the settings for running tests in the github CI - this
settings file now resides in tests/settings_ci.py

Closes: #222


sennierer:
this is not needed anymore and can be removed

PR #343, opened 2023-10-12
Ms/340 come up with a working solution for version history

### LINKED ISSUE #340: come up with a working solution for version history
this solution should also allow to introduce old versions of an entity (eg for the OEBL version of apis)
[django-simple-history](https://github.com/jazzband/django-simple-history) seems like the better way forward than django-reversion

related to https://github.com/acdh-oeaw/apis-core-rdf/issues/242

sennierer:
Simple implementation of first tests to see if the scenario we want to achieve is possible with django-simple-history:
- we should have a possibility to create new versions of data (eg a new published version of APIS data)
- from a given version of an entity we need a simple path to the correct versions of the related entities
- we want to add further data to the "version" eg a version tag and/or a comment what was changed

fixes #340


sennierer:
so far provides:
- a VersionMixin that can be used in the models to turn on versioning
- a API endpoint that serializes all the versions for a given entity_type & pk (including diffs)
- unittests (using the OEBL ontology) that test basic functionality

still missing:
- API endpoint should also return changed/added relations
- Admin integration
- templatetags for:
      - showing a track change
      - adding a version tag to given version
      - merging several changes into one version
      - undo changes
- list of diffs should provide a proper serialization of m2m relations. Currently its a simple array of the through table

apis_core/apis_history/api_views.py
```diff
@@ -0,0 +1,41 @@
...
+    serializer_class = HistoryLogSerializer
+
+    def get_queryset(self):
+        id = self.request.query_params.get("id", None)
+        model = self.request.query_params.get("entity_type", None)
+        if id is None:
+            raise Http404("No id provided")
+        if model is None:
+            raise Http404("No model provided")
+        cls = ContentType.objects.get(app_label="apis_ontology", model=model)
```
b1rger, line 21: There is `get_entity_class_of_name` in `utils.caching`

apis_core/apis_history/api_views.py
```diff
@@ -0,0 +1,41 @@
...
+        qs = cls.model_class().history.filter(id=id)
+        for q in qs:
+            id1 = q.history_id
+            ts = q.history_date
+            exclude.extend(
+                cls.model_class()
+                .history.filter(history_id__lt=id1, history_date=ts)
+                .values_list("history_id", flat=True)
+            )
+        return qs.exclude(history_id__in=exclude)
```
b1rger, line 32: I think this logic for generating the list of versions should be moved to a `apis_history.utils.get_history(model_class, id, exclude=[])` - that way it can be reused in a `View`

apis_core/apis_history/apps.py
b1rger, line 1: Maybe we can use `versioning` as a name? We already have `apis_core` as a namespace, IMO it does not really make sense to reuse that namespace as a module name prefix

apis_core/tests/__init__.py
b1rger, line 1: Maybe it makes sense to move this to `tests` dir one level up? poetry creates the `tests` dir on the same level as the module itself.

poetry.lock
b1rger, line 1: I think this `.lock` file contains a lot more changes than removing `reversion` and adding `simple_history`

apis_core/apis_entities/templates/apis_entities/edit_generic.html
b1rger, line 1: I would prefer a DjangViews first, javascript second approach - it makes maintaining the code a lot easier. If we provide a View for listing the history of an entity, we can easily include that here

sennierer:
superseded by #698

PR #422, opened 2023-11-17
fix: rewrite error message

koeaw:
The error message which can appear when trying to load entity list views has become outdated (makes references to entity settings declarations we don't use anymore), this PR fixes this + rewords the message for clarity.


b1rger:
thanks!

PR #458, opened 2023-12-06
fix deletion of entity objects / GenericEntitiesDeleteView

koeaw:


koeaw:
Update on why this is still a draft:

* After last week's discussion at the JF, which also touched on URLs, I realised I never had another look at my `urlpatterns`, so was wondering if that's what's to blame for some of what I'm (not) seeing
* I guess ideally, I wouldn't create a separate deletion confirmation for entities, even though I find it more intuitive to have them in one place (I didn't immediately find the shared one)... so I'd need to invest more time to figure out if/where one for entities differs from a more generic one, and either edit the existing one or create one that extends the shared on

koeaw:
Reworked to solve only the most pressing issue, [see this comment](https://github.com/acdh-oeaw/apis-core-rdf/issues/485#issuecomment-1864735667).

Interestingly, this actually shows the object name in the confirmation template for me now. :thinking:

b1rger (commented):
This shows the `__str__` of the object being deleted, seen as a `RootObject` - `RootObject.__str__` returns `RootObject.name`. This works, but we are also planning on getting rid of `RootObject.name`. If we instead use [`get_queryset`](https://docs.djangoproject.com/en/4.2/ref/class-based-views/mixins-single-object/#django.views.generic.detail.SingleObjectMixin.get_queryset) instead of setting the model:
```
def get_queryset(self):
   return self.entity_model.objects.all()
```
it should instead use the `__str__` of the specific entity (i.e. if the `Person` entity overrides the `__str__` (i.e. to include `first_name`) it uses that one) - could you try that?

koeaw:
Ahh, right!

Sorry, forgot that we discussed this (I think), will change it.

b1rger:
Thanks!

PR #463, opened 2023-12-07
fix: tooltips for class docstrings, whitespace issues

koeaw:
- fixes linebreaks between text and HTML tags caused by templatetag
- checks for existence of proper docstring for classes (see PEP 257) instead of falling back on class definition

References:
https://peps.python.org/pep-0257/

### DIFF
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -156,9 +156,15 @@ class GenericListViewNew(
         context[self.context_filter_name] = self.filter
         context["entity"] = self.entity  # model slug
         context["app_name"] = "apis_entities"
-        context["docstring"] = f"{model.__doc__}"
-
         context["entity_create_stanbol"] = GenericEntitiesStanbolForm(self.entity)
+
+        # model.__doc__ defaults to "ClassName(attr1, attr2)" when there is no
+        # actual docstring, which is not helpful to output on the frontend
+        if (
+            f"{model.__doc__}"
+            != f"{model.__name__}({', '.join([f.name for f in model._meta.fields])})"
+        ):
+            context["docstring"] = f"{model.__doc__}"
 
         if "browsing" in settings.INSTALLED_APPS:
             from browsing.models import BrowsConf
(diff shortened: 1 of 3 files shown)


koeaw:
That link in the bot message is broken (... and I dunno why it doesn't like the formatting of my commit message).

koeaw:
Hm, hold on, still need to check something...

koeaw:
The original PR description is outdated after a refactor and force push and/but this is still set to draft because:
* I hadn't considered the ramifications of my proposed CSS change on the placement of the "Create new" button on entity (list) pages (side note: button text could be shortened to "Create"),
* I had to give up tracking down how `model.__doc__` is constructed when there is no docstring due to time constraints, so it's possible there is an easier/more elegant way to check for & exclude the not-so-helpful fallback.

b1rger:
>    I had to give up tracking down how `model.__doc__` is constructed when there is no docstring due to time constraints, so it's possible there is an easier/more elegant way to check for & exclude the not-so-helpful fallback.

https://github.com/django/django/blob/b287af5dc954628d4b336aefc5027b2edceee64b/django/db/models/base.py#L399

koeaw:
> > I had to give up tracking down how `model.__doc__` is constructed when there is no docstring due to time constraints, so it's possible there is an easier/more elegant way to check for & exclude the not-so-helpful fallback.
> 
> https://github.com/django/django/blob/b287af5dc954628d4b336aefc5027b2edceee64b/django/db/models/base.py#L399

Thank you!

@birger I wanted to tag you anyway because while I don't consider this finished yet/was expecting to work on this more, as per my previous comment, please let me know if you have any notes/feedback (if not a proper review yet). Just had to get that earlier update out of the way.

b1rger:
> please let me know if you have any notes/feedback

I would put the logic of "calculating" the docstring into the class - either the `AbstractEntity`, but we might even use this for non entity models, so `RootObject` might make more sense. For example something like this:
```
@classmethod
@property
def description(cls):
    return cls.__doc__ 
```
That way we don't have to add it to the context, because a `ListView` has access to the model class via the queryset anyway - and in the template we can simply use `{{ object_list.model.description }}`. There is one problem though, the `class_definition` templatetag "filters" the context and only passes the two variables:

PR #470, opened 2023-12-11
feat(apis_entites)!: remove person detail template

### LINKED ISSUE #397: Get rid of `person_detail_generic.html`
We decided that we don't want to make any assumptions about the ontology - therefore we should not expect the ontology to contain a person entity. But even if the ontology contains a person entity, the generic template is generic enough to provide a detail view for persons

gythaogg:
apis_entities should not assume the presence of a "person" model as it is not present in apis_entities.

BREAKING CHANGE: any template in your ontology that inherits from the person_detail_generic template needs to point to the detail_generic template instead

closes #397


b1rger:
thanks!

Issue #506, opened 2023-12-15
Choices with enum types don't use labels but values in (some) templates


koeaw:
`Choices` fields can be implemented with [enumeration types](https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types) which allows setting both a **value** and a human-readable **label** for each choice.

We use them so we can set English defaults in the database (= values) and display project-specific German wordings (= labels) for predefined choices to our content managers. This mirrors how we handle fields, which have English field names and German verbose names.

**Describe the bug**
The _labels_ display fine in forms (create/edit views) but the detail view only shows the _values_.

The note in [this SO answer](https://stackoverflow.com/a/58051918/) / the comments below potentially describe what the problem is (?) though I haven't been able to work out why the labels work as expected in form-based views. Looked at crispy forms as well but couldn't find anything that would have explained it. Was also wondering if it could have to do with (a lack of) handling of `gettext_lazy`.

Django 5.0 slightly changed how enumeration types are handled; they are now meant to be set like this:

```
        choices=YEAR_IN_SCHOOL_CHOICES,
```

vs. how it's been done so far:

```
        choices=YEAR_IN_SCHOOL_CHOICES.choices,
```

see also [info box at the end of the section](https://docs.djangoproject.com/en/5.0/ref/models/fields/#enumeration-types) in the docs for 5.0... but I don't know if this fixes this and I'm not aware of any plans for us to switch to 5.0 soon-ish, so...

**Screenshots**

**Edit view:**

![Screenshot 2023-12-13 at 17-07-38 Title Placeholder](https://github.com/acdh-oeaw/apis-core-rdf/assets/113020169/3ff25739-7b9d-4979-b11b-28276266b47f)

**Detail view:**

![Screenshot 2023-12-13 at 17-07-23 A Figur (ID 1)](https://github.com/acdh-oeaw/apis-core-rdf/assets/113020169/796b44e5-5af7-4271-b283-038ffd6a5845)

for a model mixin which looks like this:

```
from django.utils.translation import gettext_lazy as _

class StatusMixin(models.Model):
    class ProgressStates(models.TextChoices):
        CREATED = "created", _("neu angelegt")
        ...(snip)...

    progress_status = models.CharField(
        max_length=255,
        choices=ProgressStates.choices,
        null=True,
        verbose_name=_("Bearbeitungsstatus"),
    )
    ...(snip)...
```

koeaw:
[...]

PR #507, opened 2023-12-15
refactor: display values of fields in display view

koeaw:
- use _get_FIELD_display() to set field values to fix display values of choices using enumeration types (uses choices labels over values)
- simplify check for exclude_fields by incorporating it into existing function call

Closes: #506

### DIFF
--- a/apis_core/apis_entities/detail_generic.py
+++ b/apis_core/apis_entities/detail_generic.py
@@ -126,9 +126,11 @@ class GenericEntitiesDetailView(ViewPassesTestMixin, EntityInstanceMixin, View):
         ]
         entity_settings = get_entity_settings_by_modelname(self.entity)
         exclude_fields.extend(entity_settings.get("detail_view_exclude", []))
-        for field, value in model_to_dict(self.instance).items():
-            if field not in exclude_fields:
-                relevant_fields.append((self.instance._meta.get_field(field), value))
+
+        for field_name in model_to_dict(self.instance, exclude=exclude_fields):
+            field_obj = self.instance._meta.get_field(field_name)
+            value = self.instance._get_FIELD_display(field_obj)
+            relevant_fields.append((field_obj, value))
 
         return HttpResponse(
             template.render(


koeaw:
This basically only minimally changes what's already there, not sure if it'd be preferable to get rid of the current solution altogether.

This could also possibly be refactored further to simplify that earlier `exclude_fields` calculation.

b1rger (approved):
Thanks! I did not know about `_get_FIELD_display`, that makes more sense than using the form field value, which `model_to_dict` provides. I'm a little hesitant, because its a private method, but I'm ok with that for now. That refactoring makes using `model_to_dict` in fact a bit useless and it might make sense to move that whole block to a templatetag, but lets just merge it for now to fix that bug and revisit the rest later

koeaw:
It took me a while to find out about `_get_FIELD_display()` myself while frustratingly looking for a more generic solution to `get_FOO_display()`.

I share your reservation about it being a private method, but old Django tickets I've dug up suggest there are no plans from the Django team to make functionality like this available in other ways, so... yah.

I'll note I ran into problems with _some_ fields while experimenting with getting the labels to display, but don't _quite_ remember anymore what I did/was trying to do. IIRC the problems were related to (direct or indirect?) lookups of model names, either via `_meta` or `attname` (which might be one and the same), and it happened with fields I wasn't actually aware of. Something something autofields? ... It's possible `model_to_dict` filters those out and that's part of the reason I ended up sticking with it after all (?!).

koeaw:
Ok, I think I know now: `_get_FIELD_display()` does not work for many-to-one relationships.

When I look up all fields with `self.instance._meta.get_fields()` and try to use `_get_FIELD_display()` on them, I get "'ManyToOneRel' object has no attribute 'attname'" errors for: `uri`, `triple_set_from_subj`,  `triple_set_from_obj`.

... So I'm assuming this means we won't be able to display foreign key in our display views? Which hadn't occurred to me yet, because while we've included them in our models, we haven't actually _set_ any yet.

What's the solution to this?

koeaw:
Maybe I should move my comments out of this PR and into the linked issue.

koeaw:
Oups, just found this still-open PR...

b1rger:
I'm not sure it makes sens to invest time in this - there is a WIP branch to refactor `apis_entities` to use the new `generic` app - if everything goes to plan, the whole `detail_generic.py` will be replaced

koeaw:
> I'm not sure it makes sens to invest time in this - there is a WIP branch to refactor `apis_entities` to use the new `generic` app - if everything goes to plan, the whole `detail_generic.py` will be replaced

True, true, it doesn't make sense to pursue this particular PR further -> closing.

Issue #483, opened 2023-12-12
Implement `get_member_for_entity` for filters


b1rger:
same as with tables and forms

b1rger:
This is part of the `generic` app and therefore not needed anymore

PR #510, opened 2023-12-19
feat: use `get_member_for_entity` for filterset

b1rger:
Closes: #483


b1rger:
This is part of the `generic` app and therefore not needed anymore

Issue #509, opened 2023-12-18
Make list + crud views even more general


b1rger:
It would be great if we ship those views for *all* the models, not only the entities and relations

PR #526, opened 2024-01-09
feat: introduce generic app

b1rger:
This commit introduces a `generic` app, that provides generic Create,
View, Update, Delete, List and Autocomplete views that work with
ContentTypes as well as an API viewset that also works with ContentTypes
Most of the used classes (Table, FilterSet, Serializers, ...) can be
overriden an every level of the inheritance chain.

Closes: #509

### DIFF
--- a/apis_core/generic/helpers.py
+++ b/apis_core/generic/helpers.py
@@ -1,0 +1,71 @@
+import functools
+import inspect
+import importlib
+
+from django.db.models import CharField, TextField, Q, Model
+
+
+def generate_search_filter(model, query):
+    """
+    Generate a default search filter that searches for the `query`
+    in all the CharFields and TextFields of a model (case-insensitive)
+    """
+
+    fields_to_search = [
+        field.name
+        for field in model._meta.fields
+        if isinstance(field, (CharField, TextField))
+    ]
+    q = Q()
+
+    for token in query:
+        q &= functools.reduce(
+            lambda acc, field_name: acc | Q(**{f"{field_name}__icontains": token}),
+            fields_to_search,
+            Q(),
+        )
+    return q
+
+
+def mro_paths(model):
+    """
+    Create a list of MRO classes for a Django model
+    """
+    paths = []
+    for cls in filter(lambda x: x not in Model.mro(), model.mro()):
+        paths.append(cls.__module__.split(".")[:-1] + [cls.__name__])
+    return paths
+
+
+def template_names_via_mro(model, suffix=""):
+    """
+    Use the MRO to generate a list of template names for a model
+    """
+    mro_prefix_list = ["/".join(prefix) for prefix in mro_paths(model)]
+    return [f"{prefix.lower()}{suffix}" for prefix in mro_prefix_list]
+
+
+def class_from_path(classpath):
+    """
+    Lookup if the class in `classpath` exists - if so return it,
+    otherwise return False
+    """
+    module, cls = classpath.rsplit(".", 1)
+    try:
+        members = inspect.getmembers(importlib.import_module(module))
+        members = list(filter(lambda c: c[0] == cls, members))
+    except ModuleNotFoundError:
+        return False
+    if members:
+        return members[0][1]
+    return False
+
+
+def first_match_via_mro(model, path: str = "", suffix: str = ""):
+    """
+    Based on the MRO of a Django model, look for classes based on a
+    lookup algorithm and return the first one that matches.
+    """
+    paths = list(map(lambda x: x[:-1] + [path] + x[-1:], mro_paths(model)))
+    classes = [".".join(prefix) + suffix for prefix in paths]
+    return next(map(class_from_path, classes), None)
--- a/apis_core/generic/templates/generic/generic_confirm_delete.html
+++ b/apis_core/generic/templates/generic/generic_confirm_delete.html
@@ -1,0 +1,18 @@
+{% extends basetemplate|default:"base.html" %}
+
+{% block content %}
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-body">
+        <form action="" method="post">
+          {% csrf_token %}
+          <h4>Confirm deletion of:</h4>
+          <div class="text-center">
+            <strong>{{ object }}</strong>
+          </div>
+          <input class="btn btn-danger" type="submit" value="Yes, delete" />
+        </form>
+      </div>
+    </div>
+  </div>
+{% endblock content %}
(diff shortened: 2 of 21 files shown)


koeaw:
Just had a look: works for me! (But I also realised that things I expected to show up can't actually show up because they are mixins, oups.) I like those icons in the list view!

What I'm wondering about (not having explored the code in detali yet):
* I'm guessing I can (and would) override the default URL in my own `urls.py` file?
* Are `APIS_SETTINGS` set up to also accept configurations for non-entity classes?

koeaw:
Ok, I think that's a "no" on the second question, just tried to add `table_fields` for non-entity class `DataSource` to `APIS_ENTITIES`, which didn't take. Also just noticed that the table view just shows the word **Desc** as header for the first column.

Screenshot 2024-01-17 at 11-56-32 Title Placeholder

koeaw:
Ah, but the fields all show up on the left, in the dropdown for additional table columns.

b1rger:
>     1. I'm guessing I can (and would) override the default URL in my own `urls.py` file?

I'm not sure what you mean exactly - but you *can* add additional routes to the individual views

>     2. Are `APIS_SETTINGS` set up to also accept configurations for non-entity classes?

No. The idea is to *not* use `APIS_SETTINGS` but to automaticall find what should be used - so for your example, you'd be able to create `apis_ontology.tables.DataSourceTable` and it will be used and you can design it as you like

koeaw:
> I'm not sure what you mean exactly - but you _can_ add additional routes to the individual views

What I meant was that I don't like the default URL format, `/apis/apis_ontology.lowercaseclassname/` – if adding _additional_ custom routes to get to the same views (rather than replacing the defaults) solves the issue, great!

> so for your example, you'd be able to create apis_ontology.tables.DataSourceTable and it will be used and you can design it as you like

Ahh, got it, thank you!

apis_core/generic/views.py
```diff
@@ -0,0 +1,202 @@
...
+                get_permission_codename(
+                    self.permission_action_required, self.model._meta
+                )
+            ]
+        return []
+
+
+class List(GenericModelMixin, PermissionRequiredMixin, SingleTableMixin, FilterView):
+    """
+    List view for a generic model
```
koeaw, line 64: I would recommend ending full sentences with the appropriate punctuation, i.e. full stops.

This not only helps readability overall but also prevents IDEs from flagging blocks of texts for failing language checks.

My IDE interprets _all_ the Docstrings in these files to contain run-on sentences (!) and it flagged this particular one for nonsensical grammar because it thinks the uncountable noun "access" was given an indefinite article ("a generic model access").

apis_core/generic/views.py
```diff
@@ -0,0 +1,202 @@
...
+    def get_queryset(self):
+        return (
+            first_match_via_mro(self.model, path="querysets", suffix="ListViewQueryset")
+            or self.model.objects.all()
+        )
+
+
+class Detail(GenericModelMixin, PermissionRequiredMixin, DetailView):
+    """
+    Detail view for a generic model
```
koeaw, line 113: Same IDE grammar flagging problem as above due to missing full stops.

apis_core/generic/views.py
```diff
@@ -0,0 +1,202 @@
...
+    Access requires the `<model>_view` permission
+    """
+
+    permission_action_required = "view"
+
+
+class Create(GenericModelMixin, PermissionRequiredMixin, CreateView):
+    """
+    Create view for a generic model
+    Access requires the `<model>_add` permission
```
koeaw, line 123: Same IDE grammar flagging problem as above due to missing full stops.

apis_core/generic/views.py
```diff
@@ -0,0 +1,202 @@
...
+        return reverse(
+            "apis:generic:list",
+            args=[self.request.resolver_match.kwargs["contenttype"]],
+        )
+
+
+class Update(GenericModelMixin, PermissionRequiredMixin, UpdateView):
+    """
+    Update view for a generic model
+    Access requires the `<model>_change` permission
```
koeaw, line 163: More grammar flagging due to missing full stops in Docstrings.

apis_core/generic/views.py
```diff
@@ -0,0 +1,202 @@
...
+            args=[self.request.resolver_match.kwargs["contenttype"]],
+        )
+
+
+class Autocomplete(
+    GenericModelMixin, PermissionRequiredMixin, autocomplete.Select2QuerySetView
+):
+    """
+    Autocomplete view for a generic model
+    Access requires the `<model>_view` permission
```
koeaw, line 189: More red underlines for "a generic model access".

apis_core/generic/helpers.py
```diff
@@ -0,0 +1,71 @@
...
+import inspect
+import importlib
+
+from django.db.models import CharField, TextField, Q, Model
+
+
+def generate_search_filter(model, query):
+    """
+    Generate a default search filter that searches for the `query`
+    in all the CharFields and TextFields of a model (case insensitive)
```
koeaw, line 11: case insensitive -> case-insensitive

apis_core/generic/templates/generic/generic_confirm_delete.html
```diff
@@ -0,0 +1,20 @@
...
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-body">
+        <form action="" method="post">
+          {% csrf_token %}
+          <h4>
+            Confirm deletion of:
+            <div class="text-center">
+              <strong>{{ object }}</strong>
+            </div>
```
koeaw, line 13: `<div>` tags should not be placed inside HTML headings. An inline tag which can be used in this place is `<span>`.

If the `<div>` is needed for formatting purposes, it needs to go around the `<h4>` tag.

b1rger:
> Ah, but the fields all show up on the left, in the dropdown for additional table columns.

I've created #546 for that - not sure yet if it is the right approach, though

b1rger:
> Also just noticed that the table view just shows the word **Desc** as header for the first column.

**Desc** for description - it is simply the string representation of the record - I'm happy to replace **Desc** with something else if that term leads to confusion

b1rger:
> > > I'm not sure what you mean exactly - but you _can_ add additional routes to the individual views
> > 
> > 
> > What I meant was that I don't like the default URL format, `/apis/apis_ontology.lowercaseclassname/` – if adding _additional_ custom routes to get to the same views (rather than replacing the defaults) solves the issue, great!
> 
> Yes, kind of. So you can see in `generic.views.GenericModelMixin` that the views need the `contenttype` as part of the URI - and this is generated from `app_name.model` using `generic.urls.ContenttypeConverter`. You can add any route that uses these views if you make sure the views have access to the `kwargs['contenttype']`.

I pushed a POC to 39c9f05d0463391fb8b3033df0c27411b01640e3, which demonstrates how that could look

PR #536, opened 2024-01-11
refactor: use model verbose names in property table

koeaw:
Switch from class name to verbose name, capitalise strings instead of title-casing them


b1rger:
Thanks!

Issue #586, opened 2024-01-30
Add `duplicate` button to generic table


b1rger:

b1rger:
This was implemented in 45de48bea2b81823873066820a079635b8d7a04a

PR #590, opened 2024-01-31
feat: add a duplicate button to the generic table

b1rger:
This commit introduces a `duplicate` view in the generic app, which
simply calls the `duplicate()` method of the requested object and
redirects to the `Update` view of the created object.
We also introduce a `duplicate` button in the listview table, that is
only shown if the user has the permission to create the listed object
and if the object has the `duplicate()` method.

Closes: #586

### DIFF
--- a/apis_core/generic/tables.py
+++ b/apis_core/generic/tables.py
@@ -26,40 +26,43 @@ class CustomTemplateColumn(tables.TemplateColumn):
         )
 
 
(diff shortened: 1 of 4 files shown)


b1rger:
This should be part of the entities app, not the generic app - duplication is an entities feature

b1rger:
This was superseded by 45de48bea2b81823873066820a079635b8d7a04a

PR #597, opened 2024-01-31
Add add-to-project.yml workflow based on prosnet-workflows template

koeaw:
I'm actually not _entirely_ sure if GitHub really differentiates between strings and numbers in action variables, so left the `PROJECT_NUM` setting as-is and set the `GH_PROJECT_NUM` var. :eyes:

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: 2024 K Kollmann
+# SPDX-License-Identifier: MIT
+
+name: Add repository issues to Prosnet project
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    uses: acdh-oeaw/prosnet-workflows/.github/workflows/add-to-project.yml@v0.4.0
+    secrets:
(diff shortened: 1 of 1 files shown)


koeaw:
Hmmmm, actually just saw this hint "Note: Variable values are exposed as plain text."

You didn't happen to also test the workflow with a `var` set, did you, @b1rger?

b1rger:
> Hmmmm, actually just saw this hint "Note: Variable values are exposed as plain text."
> 
> You didn't happen to also test the workflow with a `var` set, did you, @b1rger?

Nope, I did set the `PROJECT_NUM` directly (didn't see the reason for using an action variable there...)

koeaw:
> Nope, I did set the `PROJECT_NUM` directly (didn't see the reason for using an action variable there...)

Hmm, the original idea was that because projects shouldn't be accessible by non-team org members, i.e. are basically invisible to them, maybe there could be a reason for not wanting to share with others the URL either (or how many projects there are overall, or the how-many-eth project one's own is or similar). Also with regard to the workflow being adaptable to non-org projects.

But that's probably overkill for our purpose... And in the case of 3rd-party projects, one would probably actually want to obscure an account name more than a project no. (which the reusable workflow is not set up to deal with). :p

Ok, will update the file here and make a note to adapt the parent workflow/docs at a later point.

b1rger (approved):

Issue #529, opened 2024-01-10
Old jQuery code is buggy


koeaw:
**Describe the bug**
[This code here](https://github.com/acdh-oeaw/apis-core-rdf/blob/ae6f2c47c5033b1a985169d45e1095c6130a0647/apis_core/apis_entities/templates/apis_entities/apis_templatetag.html#L31-L35) doesn't work, results in `Uncaught TypeError: $.apScrollTop is not a function`, can we remove it?

I haven't been able to identify what it's supposed to do, or which plugin it belongs/belonged to, do you remember, @sennierer? 

PR #329 worked around another problem with it by duplicating the surrounding code/document-ready-check. I suggest we consolidate the two checks and get rid of the `$.apScrollTop` thing. Otherwise it needs fixing.

b1rger:
I guess its a leftover of the scroll to top javascript, which was replaced in 964a58105b99f819581d0f2014099c7e8363e8e2. I'd say it can safely be removed

PR #670, opened 2024-02-26
fix: remove buggy jQuery code

koeaw:
Closes: #529


b1rger:
Could you add `(apis_entites)` to the commit message after the fix?

koeaw:
@sennierer assigned you for review because you didn't reply to the original issue. :p

koeaw:
> Could you add `(apis_entites)` to the commit message after the fix?

Yes, can do!

b1rger:
thanks!

PR #686, opened 2024-03-01
refactor(apis_relations): drop custom table factory

b1rger:
This commit drops the `get_generic_triple_table` method that was used to
generate the table classes for the triple tables in the edit and the
detail view. For the edit and the detail view there are now the templatetags
`triple_tables_related_detail` and `triple_tables_related_edit` which
provide those tables.
The `generic_order*` and `generic_render*` methods that were located in
the `apis_metainfo.tables` module where dropped and replaced by custom
table columns `StartDateColumn` and `EndDateColumn`.


b1rger:
Superseded by various smaller commits

PR #698, opened 2024-03-04
Ms/340 add simple history to apis

### LINKED ISSUE #340: come up with a working solution for version history

sennierer:
Tests are running through again and some of the comments from #343 have been included. JavaScript and template is not yet included as I will add a htmx implementation for the views.

supersedes #343 
resolves #340


b1rger (changes requested):

apis_core/apis_history/urls.py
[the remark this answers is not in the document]
sennierer: I didnt use the Converter because the same function is used by the API where the entity comes via query param...but I will refactor the code to use the converter for the view and the contenttype framework for the api

apis_core/urls.py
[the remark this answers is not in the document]
sennierer: had that originally, but then the api route (the one not included in the router) will be in `apis/history/api` instead of `apis/api/history` ... anyway, moved it

apis_core/tests/test_simple_history.py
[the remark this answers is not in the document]
sennierer: why would we want to have the tests not in the module itself? This would mean if you install the package you actually cant run the tests? Or do I get the path wrong (but the collections tests are also on root level)?

b1rger (changes requested):

apis_core/apis_history/templates/apis_core/apis_history/apishistorytablebase_detail.html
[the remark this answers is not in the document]
sennierer: resolved by passing the `object.instance` to entity_base_nav -> with if clause in the entity_base_nav including the history button if history is installed

apis_core/apis_history/templates/apis_history/change_log.html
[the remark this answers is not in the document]
sennierer: this uses a table in a table cell and its working. I'll create an issue to improve that at a later stage

apis_core/apis_history/views.py
```diff
@@ -0,0 +1,83 @@
...
+            key=lambda x: datetime.strptime(x["timestamp"], "%d.%m.%Y %H:%M"),
+            reverse=True,
+        )
+        context["object"] = model.model_class().objects.get(id=self.kwargs["pk"])
+        context["modelname"] = model.model_class().__name__.lower()
+        context["app_label"] = model.model_class()._meta.app_label
+        return context
+
+
+def create_new_version(self, model, id):
```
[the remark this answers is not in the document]
sennierer: thats due to function based view

apis_core/apis_relations/migrations/0007_delete_historicaltriple.py
b1rger, line 1: Maybe you can remove the migrations and recreate them, it seems there is now one migration that creates `HistoricalTriple` and one that removes `HistoricalTriple`

apis_core/history/api_views.py
b1rger, line 1: Those ViewSets could use the `contenttype` converter instead of using query params. If using query params has any advantages, it could be done with `apis_core.generic.serializers.ContentTypeInstanceSerializer`

apis_core/history/templatetags/apis_history_templatetags.py
```diff
@@ -0,0 +1,31 @@
...
+    obj = context["object"]
+    return triple_sidebar_history(
+        obj.pk, obj.__class__.__name__.lower(), context["request"], detail
+    )
+
+
+@register.filter
+def opts_from_history(obj):
+    return obj.instance._meta
+
```
b1rger, line 21: filter is not being used anywhere

apis_core/history/urls.py
```diff
@@ -0,0 +1,34 @@
...
+router.register(r"temp_triple/edit_log", TempTripleHistoryLogs, "temptriplehistorylog")
+
+urlpatterns = [
+    path(
+        "change_history/<contenttype:contenttype>/<int:pk>/",
+        views.ChangeHistoryView.as_view(),
+        name="change_history",
+    ),
+    path(
+        "add_new_version/<contenttype:model>/<int:pk>/",
```
b1rger, line 24: s/model/contenttype (or `content_type`?)

apis_core/history/views.py
```diff
@@ -0,0 +1,69 @@
...
+
+class ChangeHistoryView(GenericModelMixin, SingleObjectMixin, TemplateView):
+    template_name = "history/change_history.html"
+
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        return super().get(request, *args, **kwargs)
+
+
+def create_new_version(self, model, pk):
```
b1rger, line 35: `self` should be renamed

b1rger (changes requested):

apis_core/history/views.py
sennierer: renamed to `request`

apis_core/history/urls.py
sennierer: used `contenttype` for consistency

apis_core/history/templatetags/apis_history_templatetags.py
sennierer: removed

apis_core/history/api_views.py
sennierer: removed the viewsets, made them genericviews and used contenttypes in urls

apis_core/apis_relations/migrations/0007_delete_historicaltriple.py
sennierer: recreated migrations

b1rger (approved):

PR #704, opened 2024-03-05
feat(generic): add generic RDF renderers and serializers

b1rger:
**feat(generic): allow apiview to use custom serializers & renderers**

The commit configures the generic API view to look for more differently
configured serializers. The serializers can now also be prefixed with
the requested format, as configured in the renderer.

**feat(generic): ship generic RDF renderers**

They can be enabled as described in the upstream DRF framework
documentation:
https://www.django-rest-framework.org/api-guide/renderers/#setting-the-renderers
They expect the renderer to put triples in the data object and they
build the graph based on these triples.

**feat(generic): allow to specify a dialect when requesting data**

If a dialect is either specified via the `dialect` parameter or the
`DIALECT` HTTP header, only renderer that have that dialect listed in
their `.dialects` class attribute are considered.
If no dialect is specified, only renderers that have the `.dialects`
class attribute *not* set or have `None` in the list are considered.

### DIFF
--- a/apis_core/generic/api_views.py
+++ b/apis_core/generic/api_views.py
@@ -1,6 +1,6 @@
 from rest_framework import viewsets
 from .serializers import serializer_factory, GenericHyperlinkedModelSerializer
-from .helpers import module_paths, first_member_match
+from .helpers import module_paths, first_member_match, makeclassprefix
 from .filterbackends import GenericFilterBackend
 
 
@@ -31,6 +31,12 @@ class ModelViewSet(viewsets.ModelViewSet):
         serializer_class_modules = module_paths(
             self.model, path="serializers", suffix="Serializer"
         )
+        prefix = makeclassprefix(self.request.accepted_renderer.format)
+        serializer_class_modules = (
+            module_paths(self.model, path="serializers", suffix=f"{prefix}Serializer")
+            + serializer_class_modules
+        )
+
         serializer_class = first_member_match(
             serializer_class_modules,
             getattr(renderer, "serializer", GenericHyperlinkedModelSerializer),
--- a/apis_core/generic/helpers.py
+++ b/apis_core/generic/helpers.py
@@ -78,6 +78,13 @@ def module_paths(model, path: str = "", suffix: str = "") -> list:
 
 
 @functools.lru_cache
+def makeclassprefix(string: str) -> str:
+    string = "".join([c if c.isidentifier() else " " for c in string])
+    string = "".join([word.strip().capitalize() for word in string.split(" ")])
+    return string
+
+
+@functools.lru_cache
 def import_string(dotted_path):
     try:
         return module_loading.import_string(dotted_path)
--- a/apis_core/generic/renderers.py
+++ b/apis_core/generic/renderers.py
@@ -1,0 +1,32 @@
+from rest_framework import renderers
+from rdflib import Graph
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class GenericRDFBaseRenderer(renderers.BaseRenderer):
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        g = Graph()
+        for result in data.get("results", []):
+            match result:
+                case tuple(_, _, _):
+                    g.add(result)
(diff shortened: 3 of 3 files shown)


apis_core/apis_entities/abc.py
```diff
@@ -21,6 +24,24 @@ class E21_Person(LabelBaseModel, models.Model):
     class Meta:
         abstract = True
 
+    @property
+    def node(self):
+        if self.uri_set.all():
+            return URIRef(self.uri_set.first().uri)
```
sennierer, line 30: If we stick with the serialization per attribute and keep the definition of the RDF serialization with the models - which has the disadvantage that we can have only one RDF serialization - we need to return here one canonical URI (aka the one that contains the base domain of the project), because other entities - such as the triples - need to know how to construct the correct URI from an ID.

apis_core/apis_entities/abc.py
```diff
@@ -21,6 +24,24 @@ class E21_Person(LabelBaseModel, models.Model):
...
+        return BNode()
+
+    def triple(self):
+        return (self.node, RDF.type, FOAF.Person)
```
sennierer, line 34: We can not use FOAF for the serialization and call the entity `E21_Person` which is a cidoc-crm class. We should consistently use one or the other

sennierer (commented):
Even though I like the idea that the definition of the RDF serialization is with the models, I think we need a more flexible approach which allows to have several different serializations that in the end can get rendered into different formats.
So either the possibility to configure a dictionary of serializers per model field....eg by extending the field classes for a `rdf_serializers` param and the GenericSerializer class then just runs through the fields and calls the - depending on the requested format - correct function or by sticking with dedicated serializer classes per format and model class.
The former would allow to reuse a lot of code. Eg we could have a `LabelSerializer` function that just adds a `object rdfs:label value` triple that could be reused in almost all objects. For the name field in a person we could then have a `rdf_serializers={"cidoc": [LabelSerializer, AppellationSerializer], "foaf": [LabelSerializer, FoafNameSerializer]}` where the AppellationSerializer could add the cidoc appellation and the Foaf one could use FOAF.surname.

apis_core/generic/api_views.py
```diff
@@ -31,6 +31,26 @@ def get_serializer_class(self):
         serializer_class_modules = module_paths(
             self.model, path="serializers", suffix="Serializer"
         )
+        prefix = makeclassprefix(self.request.accepted_renderer.format)
+        serializer_class_modules = (
+            module_paths(self.model, path="serializers", suffix=f"{prefix}Serializer")
+            + serializer_class_modules
+        )
+
+        if dialect := self.request.query_params.get("dialect"):
```
sennierer, line 40: given that `dialect` can take any string and uses the inheritance search it basically silently fails and returns just the basic serialization. I think we should:

- add the dialect as a search param in one of the basic `FilterClasses` and create an enum of allowed values similar to the search through class inheritance chain down below. Like that the allowed/configured dialects would end in the swagger definition
- and raise a proper error when the search down below doesnt find any serializer classes with the dialect name in it. Like that people get what they requestet or they get an error (but not something they didnt expect)

sennierer (changes requested):

apis_core/generic/api_views.py
b1rger: Oke, I took a little time to play around with this. If we want to have `dialect` as a parameter with enum values for the API, this basically comes down to adding a `ChoiceFilter` to the filterset class that is used for the API. We can not add this `ChoiceFilter` to the `GenericFilterSet`, because this is also used for the HTML filters in the list view.

We could inject the `ChoiceFilter` into the used filterset in the `GenericFilterBackend` *after* the filterset is chosen. But the users still have to define the choices somewhere. So why not let it be the responsiblity of the users and let them simply create a custom `ApiFilterSet` and add the `ChoiceFilter` themselves?

sennierer (approved):

sennierer:
decided to leave aside the dialect idea and stick with proper content negotiation only. We will add new custom content types for the RDF versions

PR #720, opened 2024-03-15
fix(apis_metainfo): drop RootObject.deprecated_name

b1rger:


sennierer:
decided to merge the PR in the next version

koeaw:
I just wanted to note that PR #710 includes another important change that's been missing, which continues to be relevant even after the field gets removed + a refactoring we've previously discussed.

koeaw:
Which honestly makes me think it would be simpler to just accept that PR for now instead of hurrying towards dropping the field.

b1rger:
> these changes should not be forgotten _again_.

FTR: you promised to rebase #630 so that it can be merged. That did not happen. I was asked multiple times when the release with those changes will be ready. Given that the PR had massive merge conflicts, the only way forward to fix this, was to cherry pick your changes **by hand**. I'm sorry not everything of those changes made it, when I created a new PR so that the changes could be merged.

b1rger:
> Which honestly makes me think it would be simpler to just accept that PR for now instead of hurrying towards dropping the field.

We are not hurrying at all. As you can see, the PR was created nearly a week ago and we waited to discuss this at the JFX. I'm happy to wait another week, if you want to discuss this again. But I'm not going to merge something that will be reverted in the commit right after.

koeaw:
I'm not actually here to discuss how the changes made it into the codebase, but for completeness' sake: I _did_ rework everything (_again_), but as I was running into the same conflicts on my end, it took longer than expected. When I was ready to submit my changes, they had already been merged.

koeaw:
And I don't really get what the point about the creation date of this PR is meant to achieve. I created that follow-up PR two weeks ago, immediately after noticing what was missing. ?!

PR #750, opened 2024-03-25
feat(generic): implement custom schema generator

b1rger:
The CustomSchemaGenerator is meant as a drop in replacement for the
"DEFAULT_GENERATOR_CLASS" of drf-spectacular. You can use it like this:
SPECTACULAR_SETTINGS["DEFAULT_GENERATOR_CLASS"] = 'apis_core.generic.generators.CustomSchemaGenerator'

Reasons:
The first reason is, that we are using a custom converter
(`apis_core.generic.urls.ContenttypeConverter`) to provide access to views
and api views of different contenttypes. The default endpoint generator of
does not know about this converter and therefore sees the endpoints using
this converter as *one* endpoint with one parameter called `contenttype`.
Thats not what we want, so we have to do our own enumeration - we iterate
through all contenttypes that inherit from `GenericModel` and create
schema endpoints for those programatically.
The second reason is, that the autoapi schema generator of DRF Spectacular
and our `apis_core.generic.api_views.ModelViewSet` don't work well together.
Our ModelViewSet needs the dispatch method to set the model of the
`ModelViewSet`, but this method is not being called by the generator while
doing the inspection, which means the `ModelViewSet` does not know about the
model it should work with and can not provide the correct serializer and filter
classes. Therefore we create the callback for the endpoints by hand and set
the model from there.

### DIFF
--- a/apis_core/generic/generators.py
+++ b/apis_core/generic/generators.py
@@ -1,0 +1,71 @@
+from django.contrib.contenttypes.models import ContentType
+from django.urls import reverse, resolve
+from drf_spectacular.generators import SchemaGenerator
+from drf_spectacular.generators import EndpointEnumerator
+
+from apis_core.generic.abc import GenericModel
+
+
+# Custom Schema Generator
+#
+# The CustomSchemaGenerator is meant as a drop in replacement for the
+# "DEFAULT_GENERATOR_CLASS" of drf-spectacular. You can use it like this:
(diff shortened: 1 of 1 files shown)


sennierer (approved):
Tested this with the ÖBL ontology and this works really nice.
Also tested it with overwriting the Person serializer (worked as expected) and changing the `PersonFilterSet`. Worked as expected and changes resulted in changes of the swagger.

PR #752, opened 2024-03-26
feat(history): adds tests for history plugin

sennierer:
currently tests for the history plugin lack an ontology for running. This is a draft PR until we added an ontology to the testing env.

### DIFF
--- a/apis_core/history/tests/test_simple_history.py
+++ b/apis_core/history/tests/test_simple_history.py
@@ -1,0 +1,132 @@
+from datetime import datetime, timedelta
+
+from django.contrib.contenttypes.models import ContentType
+from django.test import TestCase
+
+from apis_core.apis_relations.models import Property, TempTriple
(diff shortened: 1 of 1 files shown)


b1rger (changes requested):
We will have to move the import

sennierer:
> We will have to move the import

is moved

b1rger:
This should now be updated for the `sample_project` ontology

b1rger:
ping

sennierer:
adapted the existing tests to the sample project. presumably the check in line 11 (for the CI env variable) can be removed as the tests should now run in CI as well. However, i suggest to merge these simple tests first and lift the check for CI in another PR

b1rger (changes requested):
please change the commit message to `test (history): ...` - if we cut a release, `feat` leads to a version bump

sennierer:
good point, changed that

b1rger (approved):

PR #755, opened 2024-03-28
fix(generic): select shown columns in column filter

### LINKED ISSUE #735: change `Columns` field in list filters to also allow for deselecting columns
currently the multiselect form field shows all attributes of a model, but those that are already shown in the default table are not selected and can thus not be removed from the table.
All possible columns should be in the form and those shown should be selected, if they are unselected they should be removed from the table.

sennierer:
Handing through the table class to the filterset might be a more elegant solution than calling the mro resolver again, but given that this is cached anyway its the easier solution.

resolves #735


sennierer:
just saw that this fills in the initial values of the columns correctly, but de-selecting a default column does not remove it

PR #758, opened 2024-03-28
fix: fix css for select2 autocompletes

### LINKED ISSUE #330: css of autocompletes is faulty
in autocompletes the selection and the default text are not in the middle of the line
screenshot-127 0 0 1_8001-2023 10 10-11_40_25
setting `line-height: initial` for the span that holds the autocomplete fixes that
screenshot-127 0 0 1_8001-2023 10 10-11_42_15
-> there might be a better fix though

sennierer:
please note: this is a preliminary fix and should be removed when moving to bootstrap5, see the comments in the corresponding issue

resolves #330


koeaw:
Does VSCode support TODOs? It's a built-in in PyCharm and useful to prefix comments with which are meant to be temporary or to mark something as "to be revisited" etc.

PR #769, opened 2024-04-08
fix(relations): Property fields unicode normalization

koeaw:
- Set default value for missing name_reverse field before attempting unicode normalisation
- Run unicode normalisation on string representations of the two name fields


b1rger:
thanks!

PR #823, opened 2024-05-10
fix(history): implement history_date fallback directly in property

b1rger:
Setting the history_date fallback directly in the property lets us drop
the `save()` override and the `pre_delete` signal, because those only
implemented a check if the history_date was None.


sennierer (approved):
tested with unit tests and manually. fixes the issue with the empty history date column

PR #830, opened 2024-05-13
refactor!: drop old api views

b1rger:
We do now have a generic API view that also comes with a schema
generator and custom filters, so we don't need the old api views
anymore.


sennierer:
given that no plroject is currently actively using the old API, we can drop it

PR #835, opened 2024-05-14
fix(history): returning triples makes only sense for RootObjects

b1rger:


apis_core/history/models.py
```diff
@@ -62,6 +63,9 @@ def get_triples_for_version(
         If only_latest is True, only the latest version of a triple is returned."""
         from apis_core.apis_relations.models import TempTriple
 
+        if not isinstance(self, RootObject):
+            return []
```
sennierer, line 67: I think this should be changed to 
```python
if not isinstance(self.instance, RootObject):
```
`self` is of type eg `VersionPerson` and does not inherit from `RootObject`, `self.instance` is the actual class (eg `Person`) and thus inherits from `RootObject`

sennierer (changes requested):

sennierer (approved):

PR #841, opened 2024-05-14
fix names, job references in Ruff workflows, rename workflow file

koeaw:


koeaw:
Ah, hmm, it didn't occur to me that the job name `black`  may have been chosen on purpose (?! for better differentiation?). But if so, it should probably be changed to something less ambiguous.

Also, should the older workflow be updated to the same newer version of the prosnet workflow?

b1rger (approved):

PR #842, opened 2024-05-15
Add ÖBL JS for map, fix problems with code

koeaw:


b1rger (changes requested):
please squash the commits, no need to commit buggy code and then commit two fixes on top.
can we get rid of the use of jquery?

I think it would make sense to move the javascript part to a separate `.js` file

koeaw:
>  can we get rid of the use of jquery?

All in favour of this suggestion in general, but would postpone it to a follow-up issue.

> I think it would make sense to move the javascript part to a separate `.js` file

Yes, I was actually going to ask if we could please clean up the chaos in some of the templates by dealing with JS and CSS separately (may have mentioned this somewhere before).

koeaw:
@b1rger Commits squashed down into one.

sennierer:
Include changes made in pfp-oebl repo in this fix

b1rger:
ping

Issue #860, opened 2024-05-23
"Entities" menu shows entities in wrong order


koeaw:
Follow-on from #799: entities verbose name capitalisation was fixed, but entities continue to not be listed alphabetically (by `verbose_name`) anymore.

Source of the menu is the template tag [`entities_content_types`](https://github.com/acdh-oeaw/apis-core-rdf/blob/8da1372b031279e4aa054c4f87d223b156d1b8e3/apis_core/apis_entities/templatetags/apis_templatetags.py#L40), which returns a list of entity content types -> entities appear in consecutive order of `content_type` ids. The verbose name is added only later, in the [template](https://github.com/acdh-oeaw/apis-core-rdf/blob/8da1372b031279e4aa054c4f87d223b156d1b8e3/apis_core/core/templates/base.html#L115), via template tag [`model_meta`](https://github.com/acdh-oeaw/apis-core-rdf/blob/8da1372b031279e4aa054c4f87d223b156d1b8e3/apis_core/core/templatetags/apiscore.py#L24).

For something like e.g. [`dictsort`](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#dictsort) to work in the template, the model meta data would have to get included at an earlier point.

b1rger:
What would be the "right" order?

koeaw:
> What would be the "right" order?

Alphabetical, by verbose name (or, now: plural verbose name), like before.

koeaw:
Project-specific custom ordering could be a new, extra feature, I guess.

PR #865, opened 2024-05-27
fix(apis_entities,core): add templatetag for generating entitymenu

b1rger:
This introduces `entities_verbose_name_plural_listview_url` templatetag,
which returns all entities verbose names together with their list uri,
sorted in alphabetical order.
This is then used in the core template for the main entities menu.

Closes: #860


koeaw (approved):
This works great, puts the sort order right back as expected, thank you!

Issue #602, opened 2024-02-01
Create a dockerimage


b1rger:
we have the `base-container`, but thats specific to our needs - it would be nice to have something more targeted at the general audience. ideally it can be reused in the `base-container`

Issue #876, opened 2024-05-29
Drop `apis:` url namespace


b1rger:
APIS projects usually have included the `apis_core.urls` with setting the namespace to `apis`. This led to us using the namespace in the `apis_core` codebase. We should replace all those namespace prefixes so that projects using apis_core don't have to set a specific namespace.

PR #877, opened 2024-05-29
feat: add Docker image + sample_project entrypoint & setup

b1rger:
Closes: #602

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM python:3.11
+
+# https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE
+# Prevents Python from writing .pyc files to disk
+ENV PYTHONDONTWRITEBYTECODE 1
+
+# ensures that the python output is sent straight to terminal (e.g. your container log)
+# without being first buffered and that you can see the output of your application (e.g. django logs)
+# in real time. Equivalent to python -u: https://docs.python.org/3/using/cmdline.html#cmdoption-u
+ENV PYTHONUNBUFFERED 1
+# don't cache downloaded pip packages
+ENV PIP_NO_CACHE_DIR=1
+# disable version check of pip
+ENV PIP_DISABLE_PIP_VERSION_CHECK=1
+
+COPY . /app
+
+RUN pip install /app
+
+EXPOSE 8000
+
+USER nobody
+
+ENTRYPOINT ["/app/sample_project/entrypoint.sh"]
--- a/sample_project/entrypoint.sh
+++ b/sample_project/entrypoint.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+/app/manage.py migrate
+/app/manage.py setup || true
+/app/manage.py loaddata sample_project
+/app/manage.py runserver 0.0.0.0:8000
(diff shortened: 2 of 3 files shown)


sennierer (changes requested):
with `apis_core.urls` added to `/apis` this is not working out of the box. In the `sample_project` we should include `apis_core.urls` as the root: `urlpatterns = [path("", include("apis_core.urls", namespace="apis"))]` instead.
Also with the current setup creation of any links in the sample_project fails due to https://github.com/acdh-oeaw/apis-core-rdf/issues/884
So we should either rename the sample_project (at least in the docker-image) or fix #884 before merging this.

b1rger:
> with `apis_core.urls` added to `/apis` this is not working out of the box. In the `sample_project` we should include `apis_core.urls` as the root: `urlpatterns = [path("", include("apis_core.urls", namespace="apis"))]` instead.

I think we should fix #876 instead. Including the apis URIs in a specific namespace should not be a requirement for a running instance.

sennierer:
makes sense. But my comment was regarding the path where `apis_core.urls` is currently included in the sample_project. Given that it is included under `/apis` there is no url resolving under `/` and the login is not correctly redirecting. So no matter if we get rid of the apis namespace, the path of the url include should be the root instead of `/apis`.

b1rger:
> But my comment was regarding the path where `apis_core.urls` is currently included in the sample_project. Given that it is included under `/apis` there is no url resolving under `/` and the login is not correctly redirecting.

Ah, oke, thanks for clarifying, then we are on the same page, that it should work with whereever we include `apis_core.urls` under (be it "", "/apis", "/foobar" or whatever).

I can not reproduce the login not redirecting correctly - how did you login?

(I can reproduce the logout not redirecting correctly, though - this seems to be [this problem](https://stackoverflow.com/questions/77928535/django-can-login-but-cant-logout-405-method-not-allowed), although I'm not sure why that did not come up yet)

sennierer:
when the url is at "/apis", you are not logged in and you click on lets say "Person" (to go to the list of persons) it redirects to /accounts/login/ but the login page is actually under "/apis/accounts/login/"

b1rger:
> when the url is at "/apis", you are not logged in and you click on lets say "Person" (to go to the list of persons) it redirects to /accounts/login/ but the login page is actually under "/apis/accounts/login/"

Ah, yes, indeed! But I'm not sure if we can make this run out of the box. `/accounts/login` is [the default](https://docs.djangoproject.com/en/4.2/ref/settings/#login-url) and it assumes that the user is including `django.contrib.auth.urls` under `/accounts` - which we do in most projects **in addition** to the apis shipped route `/apis/accounts`. So there are two approaches: a) use the apis shipped route and set your `LOGIN_URL` accordingly b) include `django.contrib.auth.urls` yourself

sennierer:
ok, this solves the accounts urls problem (one or the other way), but still the root url of the sample_project doesnt serve anything. So why not just change the path from `path("apis/", include("apis_core.urls", namespace="apis"))` to `path("", include("apis_core.urls", namespace="apis"))` as suggested before. Like that just starting the docker container and pointing the browser to "http://localhost:8000" shows the default APIS landing page and the login works out of the box.
Its just the sample_project. Main goal should be to run right away.

sennierer (approved):
Works nicely, apart from the fact that when properties are added, server needs to be restarted to actually get the relation forms. Given that this is another issue I opened #914

PR #886, opened 2024-06-04
fix: change redirect in uri resolver to new generic templates

### LINKED ISSUE #885: URI resolver redirects to detail-pages that no longer exist
error: `Reverse for 'person-detail' not found. 'person-detail' is not a valid view function or pattern name.`
request: `/apis/api/metainfo/uritoobject/?uri=https://d-nb.info/gnd/119165457`

sennierer:
resolves #885


b1rger (changes requested):
Shouldn't this redirect to the API view of the entity? Thats why we pass the `params` along...

b1rger:
was implemented in 24f91f1ae3e8f88a7cfbffcbc21de16e34246b97

PR #887, opened 2024-06-04
refactor(apis_entities): rewrite create_default_uri

b1rger:
This cleans up the create_default_uri signal receiver - mainly to drop
the dependency on `caching`, but also to make it more readable.

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -4,11 +4,10 @@ from django.contrib.contenttypes.models import ContentType
 from django.conf import settings
 from django.db.models.signals import post_save
 from django.dispatch import receiver
-from django.urls import reverse
+from django.urls import reverse, NoReverseMatch
 from django.db.models.query import QuerySet
 
-from apis_core.utils import caching
-from apis_core.apis_metainfo.models import RootObject
+from apis_core.apis_metainfo.models import RootObject, Uri
 from apis_core.apis_relations.models import TempTriple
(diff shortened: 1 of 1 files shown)


apis_core/apis_entities/models.py
```diff
@@ -199,22 +198,10 @@ def get_serialization(self):
...
-                base1,
-                reverse("GetEntityGenericRoot", kwargs={"pk": instance.pk}),
-            )
-            uri2 = Uri(uri=uri_c, domain="apis default", root_object=instance)
-            uri2.save()
+def create_default_uri(sender, instance, created, raw, using, update_fields, **kwargs):
+    if getattr(settings, "CREATE_DEFAULT_URI", True):
+        if isinstance(instance, AbstractEntity) and created:
+            base = BASE_URI.strip("/")
+            route = reverse("GetEntityGenericRoot", kwargs={"pk": instance.pk})
```
sennierer, line 205: Given that `GetEntityGenericRoot` needs to be implemented by the user, but another route within `apis_core` `GetEntiteGeneric` is automatically there. I suggest the following change:

```python
            try:
                route = reverse("GetEntityGenericRoot",
                                kwargs={"pk": instance.pk})
            except NoReverseMatch:
                route = reverse("apis_core:GetEntityGeneric",
                                kwargs={"pk": instance.pk})
```

sennierer (changes requested):

sennierer (approved):

PR #894, opened 2024-06-04
refactor: remove leftover print statements

koeaw:
These print statements look like they are leftovers from testing/debugging things during local development and could/should thus be cleaned up.


koeaw:
The ones I've left appear in `except` clauses.

b1rger (approved):

Issue #893, opened 2024-06-04
Add visual indicator for duplicate(d) view


koeaw:
After successful duplication of an object, the user is directed to the new object's default view.

It'd be helpful if there were a visual indicator to let the user know the operation was successful and that they are indeed looking at the new object.

Maybe there could even be a (temporary) backlink to the source object so they can navigate between the two while editing the new object.

PR #896, opened 2024-06-05
feat(apis_entities): add info message to duplicated object

b1rger:
Closes: #893

### DIFF
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -1,6 +1,8 @@
+from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
+from django.utils.html import format_html
 from django.views import View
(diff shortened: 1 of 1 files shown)


koeaw:
I really like this solution with the message bar at the top!

I'm trying to think if there is a way to reword the message a bit to make it even clearer that the current object is the duplicated one. I was wondering why there was only a link to the original object (very helpful!), not the duplicated one, then realised that d'oh, that's the one I'm looking at. :p

koeaw:
> I'm trying to think if there is a way to reword the message a bit to make it even clearer that the current object is the duplicated one.

Maybe the name of the current object (the duplicate) could be cut altogether, since it's shown right below the green bar anyway, and the message could be shortened to one of (in order from most to least potentially ambiguous):

> [X](#) was duplicated to:
> [X](#) was duplicated to here:
> [X](#) was successfully duplicated to the current object:

koeaw (approved):
:+1::+1::+1:

PR #903, opened 2024-06-05
fix(apis_metainfo): handle unique fields in the duplication logic

b1rger:
Up until now we hardcoded handling the `id` field, which ususally is a
primary key. Now the duplicate method handles all the unique fields and
removes them before duplicating.

### DIFF
--- a/apis_core/apis_metainfo/models.py
+++ b/apis_core/apis_metainfo/models.py
@@ -13,6 +13,10 @@ from apis_core.utils import caching, rdf
 
 from apis_core.apis_metainfo import signals
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 
 NEXT_PREV = getattr(settings, "APIS_NEXT_PREV", True)
 
@@ -48,7 +52,12 @@ class RootObject(GenericModel, models.Model):
         # but we are working with abstract classes,
         # so we have to do it by hand  using model_to_dict:(
         objdict = model_to_dict(self)
-        objdict.pop("id")
+
+        # remove unique fields from dict representation
+        unique_fields = [field for field in self._meta.fields if field.unique]
+        for field in unique_fields:
+            logger.info(f"Duplicating {self}: ignoring unique field {field.name}")
+            objdict.pop(field.name, None)
 
         # remove related fields from dict representation
         related_fields = [


koeaw:
This PR is missing the commit that fixed the duplication error (#899), but even when rebased to the latest `main`, it only fixes duplication for models with fields that are `unique` but allowed to be `NULL` (which is not the default).

For models with unique fields which are missing `null=True`, duplication continues to result in an `IntegrityError`.

Would it be possible to simply disallow duplication in that case? And to translate that to the UI, e.g. via a greyed-out/less contrasty purple duplication icon + explanatory hover text, say, "Duplication not possible due to fields with unique constraints" or similar.

koeaw:
Actually, it seems like such a model can be duplicated _once_ if none of the existing objects have their unique field set to the empty string yet.

I'll check in the DB but that's what seems to happen, the unique field gets set empty. (Which only works the one time.)

b1rger:
Hm, oke, i'll try to figure out how to find out *all* unique fields

b1rger:
Actually, it makes sense that there is an `IntegrityError`- if you set the field to not be nullable, there has to be some kind of default value. This doesn't really make any sense logicwise, to have a default value on a unique field, but thats how it is. And *if* there is a default value, that one gets used if the field is `null`- which is what happens if you duplicate an object but omit the fields that are unique.

koeaw:
> Actually, it makes sense that there is an `IntegrityError`- if you set the field to not be nullable, there has to be some kind of default value. This doesn't really make any sense logicwise, to have a default value on a unique field, but thats how it is. And _if_ there is a default value, that one gets used if the field is `null`- which is what happens if you duplicate an object but omit the fields that are unique.

Ohh. I see, so that's actually an incorrectly implemented `unique` field, then.

In which case the current solution should be fine? I'll test it once more with a properly configured `unique`-but-not`NULL` field.

koeaw:
Ok, after discussion elsewhere: setting a `default` value is irrelevant for testing this further.

This works for the (non-default) case when a `unique` field is also allowed to be `null` for now, and other cases will need to be handled in other ways, e.g. as suggested earlier via the UI.

koeaw (approved):

PR #906, opened 2024-06-05
refactor(apis_metainfo,apis_relations): replace use of custom cache

b1rger:


sennierer (approved):
saving an ajax form cant be tested (using apis_core only) until caching is gone, but I cant see why this shouldnt work.

Issue #912, opened 2024-06-05
Improve button labels on overview page


koeaw:
The button labels on the [`overview/` page](https://github.com/acdh-oeaw/apis-core-rdf/blob/dce75d90793d14f2443cd3f59cd758c7b1487a1c/apis_core/generic/templates/generic/overview.html) are based on the [`contenttypes` template tag](https://github.com/acdh-oeaw/apis-core-rdf/blob/dce75d90793d14f2443cd3f59cd758c7b1487a1c/apis_core/generic/templatetags/apisgeneric.py#L37).

On my end, they look like this:
![Screenshot 2024-06-05 at 18-07-34 Title Placeholder](https://github.com/acdh-oeaw/apis-core-rdf/assets/113020169/c64b1150-5633-43d0-8119-8b571174c9e0)

Problems:
* the ones for the new `history` tables aren't super helpful because they all have the same label, "apis_ontology | Version" 
* the verbose names are inconsistently cased (I'm assuming the `history` module uses capitalisation; Django recommends lowercasing them)
* they appear to be in random order

Maybe we can come up with better labelling while also keeping the `app_label` for better differentiation.

koeaw:
Maybe the page could be split into sections to separate the buttons for the different apps. Which would then allow to leave the app labels off the individual buttons.

b1rger:
Related: #920

PR #921, opened 2024-06-06
feat(generic): list all generic models on overview page

b1rger:
This commit introduces the templatetag `genericmodel_content_types`
which lists all content types of models that inherit from
`apis_core.generic.abc.GenericModel`. This templatetag is then used
in the overview template to create an overview of *all* those
contenttypes and not only `apis_ontology` and `collections` as it was
hardcoded beofre.

Closes: #912

### DIFF
--- a/apis_core/generic/templates/generic/overview.html
+++ b/apis_core/generic/templates/generic/overview.html
@@ -5,18 +5,17 @@
   <div class="card">
     <div class="card-header">Overview</div>
     <div class="card-body text-center">
-      {% contenttypes "apis_ontology" as contenttypes %}
-      {% for contenttype in contenttypes %}
-        <a href="{% url 'apis_core:generic:list' contenttype %}">
-          <button type="button" class="btn btn-outline-dark m-2">{{ contenttype }}</button>
-        </a>
-      {% endfor %}
-      <hr>
-      {% contenttypes "collections" as contenttypes %}
-      {% for contenttype in contenttypes %}
-        <a href="{% url 'apis_core:generic:list' contenttype %}">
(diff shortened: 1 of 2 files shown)


koeaw:
So, what I find weird about this is that the history models appear under separate, second sections for the apps they belong to, further down the page.

I think it would make more sense if they appeared in closer proximity, and under the same section headline. Ideas: the sections could be split in half, with the left side containing the main models, the right side the histories. Or, models and histories could come in pairs? First button links to main models, second to that model's history.

I'm trying to think how I'd most likely look up/look for histories, and I think in my head I'd go by model name first, history second, so the idea of button pairs seems like it would take the least amount of time searching for the right button to click on.

koeaw:
Hmm, also, I see myself opening both the model and its history in parallel , e.g. to check I'm looking at the right object if there are several similar/same-ish named ones (which is the case in Frischmuth).

b1rger:
> So, what I find weird about this is that the history models appear under separate, second sections for the apps they belong to, further down the page.

Can you create a screenshot? The PR only splits the models by the app they belong to and the versions are part of the same app, so they should be part of the same section

koeaw:
This is what it looks like on my end:

Screenshot 2024-06-24 at 12-43-48 Title Placeholder

b1rger:
> This is what it looks like on my end:
> 

Hm, interesting, thats not how it should look

b1rger:
> So, what I find weird about this is that the history models appear under separate, second sections for the apps they belong to, further down the page.

Thats fixed now

koeaw:
Yep, looks fine for me now!

Screenshot 2024-06-24 at 13-27-43 Title Placeholder

koeaw (approved):

Issue #874, opened 2024-05-28
Uri list view sorting incomplete / buggy


koeaw:
The default table for Uris (`/apis/apis_metainfo.uri/`) is not sortable by either **Entity Type** or the first column (labelled **Desc**) which shows the Uri itself:

![Screenshot 2024-05-28 at 15-10-38 Title Placeholder](https://github.com/acdh-oeaw/apis-core-rdf/assets/113020169/59058938-f895-405e-a6f3-9b1469e6bf0d)

and sorting by **related Entity** results in an error:

```
FieldError at /apis/apis_metainfo.uri/

Cannot resolve keyword 'entity' into field. Choices are: domain, id, loaded, loaded_time, rdf_link, root_object, root_object_id, uri
```

A workaround for the default column not being sortable seems to be to add  **Uri** from the additional columns dropdown on the left (values look identical?):

![Screenshot 2024-05-28 at 15-13-02 Title Placeholder](https://github.com/acdh-oeaw/apis-core-rdf/assets/113020169/3d0edbaa-f1bb-4f91-8117-da3636c45161)

sennierer:
only those fields represented in SQL tables (the queryset) are sortable. Desc is created by the `str()` function on the instance.
we clean up the table and make the entity type sortable.

sennierer:
remove Desc from URI table and add Uri column to default view.

PR #926, opened 2024-06-10
fix(apis_metainfo): make entity_type orderable in the UriTable

b1rger:
... and replace `desc` with `uri` (making it also orderable)

Closes: #874


koeaw (approved):
This works great, thx!

I have a few follow-up suggestions for how the view can be improved further but will put them in a new ticket.

PR #957, opened 2024-06-13
feat(apis_entities): add a default search field to the entity filtersets

b1rger:
This default search uses `generic.helpers.generate_search_filter` to
either search through all the Char and Textfields or through the fields
the model has defined in `_default_search_fields`.
It uses the same logic to list the searched fields in the help text of
the searchfield.

### DIFF
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -90,6 +90,37 @@ class PropertyFilter(django_filters.ModelChoiceFilter):
         return queryset
 
 
+class ModelSearchFilter(django_filters.CharFilter):
+    """
+    This filter is a customized CharFilter that
+    uses the `generate_search_filter` method to
+    adapt the search filter to the model that is
+    searched.
+    It also extracts sets the help text based on
+    the fields searched.
+    """
+
+    def __init__(self, *args, **kwargs):
+        model = kwargs.pop("model", None)
+        super().__init__(*args, **kwargs)
+
+        if model is not None and "help_text" not in self.extra:
+            if hasattr(model, "_default_search_fields"):
+                fields = model._default_search_fields
+            else:
+                modelfields = model._meta.fields
+                fields = [
+                    field.name
+                    for field in modelfields
+                    if isinstance(field, (models.CharField, models.TextField))
+                ]
+            fields = ", ".join(fields)
+            self.extra["help_text"] = f"Search in fields: {fields}"
+
+    def filter(self, qs, value):
+        return qs.filter(generate_search_filter(qs.model, value))
+
+
 def related_entity(queryset, name, value):
     entities = get_entity_classes()
     q = Q()
@@ -144,6 +175,11 @@ class AbstractEntityFilterSet(GenericFilterSet):
         self.filters["related_property"] = PropertyFilter(
             label="Related Property", model=getattr(self.Meta, "model", None)
         )
+
+        if model := getattr(self.Meta, "model", False):
+            self.filters["search"] = ModelSearchFilter(model=model)
+            self.filters.move_to_end("search", False)
+
         if "apis_core.history" in settings.INSTALLED_APPS:
             self.filters["changed_since"] = django_filters.DateFilter(
                 label="Changed since",


apis_core/apis_entities/filtersets.py
```diff
@@ -22,6 +22,30 @@
...
+                fields = model._default_search_fields
+            else:
+                modelfields = model._meta.fields
+                fields = [
+                    field.name
+                    for field in modelfields
+                    if isinstance(field, (models.CharField, models.TextField))
+                ]
+            fields = ", ".join(fields)
+            self.extra["help_text"] = f"Search in the fields: {fields}"
```
koeaw, line 43: I'd change this to `f"Search in fields: {fields}"`

koeaw:
This is really interesting!

Having tested it on Frischmuth, which already has a `search` field/filter for several entities, I'm wondering about a few things:
* Is there a way to turn this off? Or could there be? (on a per-model basis, or altogether) Or could this be opt-in rather than opt-out?
* Could this be modified so a custom field label and `help_text` could be provided? (in our case, we'd want to provide our own German strings)
* The list of field names would ideally be made up of verbose names (to match the labelling in the rest of the form/elsewhere)
* This (unexpectedly) overrides Frischmuth's custom filters because we use the same field/filter name, `search`. Could the hierarchy of filters be changed so project-specific filtes trump this generic search filter? And/or could the name be changed to something less likely to clash with custom filters (i.e. something less generic)?

Scenarios for the last point: A project wants to make use of this default search filter across models, but would need to override it in a few cases with a custom definition. Or: a project has its own primary `search` filter per model, but would _additionally_ like to make the generic search filter available.

E.g. in Frischmuth, we want our main "Search" field to be limited to very specific fields + use fuzzy search. But, knowing this generic filter exists, I think I'd like to make it available as well, as secondary, less "precise" filter, in case that's useful.

b1rger:
>   Is there a way to turn this off? Or could there be? (on a per-model basis, or altogether) Or could this be opt-in rather than opt-out?

You can turn this off by providing your own filterset - I think a search field should be part of the default filterset.

>  Could this be modified so a custom field label and `help_text` could be provided? (in our case, we'd want to provide our own German strings)

Yes, you can simply inherit from `ModelSearchFilter` and override the `help_text` in the init

>  The list of field names would ideally be made up of verbose names (to match the labelling in the rest of the form/elsewhere)

ack

>   This (unexpectedly) overrides Frischmuth's custom filters because we use the same field/filter name, `search`. Could the hierarchy of filters be changed so project-specific filtes trump this generic search filter? And/or could the name be changed to something less likely to clash with custom filters (i.e. something less generic)?

This is simply how object inheritance works in Python. You can override the `__init__` to make your custom search be the one thats picked. We can not put the default search into the filterset as a class variable, because there we don't have access to the `Meta.model`.


> Scenarios for the last point: A project wants to make use of this default search filter across models, but would need to override it in a few cases with a custom definition. Or: a project has its own primary `search` filter per model, but would _additionally_ like to make the generic search filter available.

Thats possible using object inheritance


> E.g. in Frischmuth, we want our main "Search" field to be limited to very specific fields 

Thats exactly what this search does

> + use fuzzy search.

As already discussed, this can be implemented in the `generic_search_filter` helper

> But, knowing this generic filter exists, I think I'd like to make it available as well, as secondary, less "precise" filter, in case that's useful.

b1rger:
> > E.g. in Frischmuth, we want our main "Search" field to be limited to very specific fields
> 
> Thats exactly what this search does

(... once #946 gets reviewed and merged)

koeaw:
When I last tested this (end of July), I eventually managed to work around this change clashing with/overwriting the custom `search` fields already in place in Frischmuth.

I.e. this worked for me – with limitations.

What I'm missing:
* `verbose_names` instead of actual field names for the list of fields in the search field's label
* `ArrayFields` which have `CharFields` or `TextFields` as `base_field` being covered as well
*  (?*) `ManyToManyFields` which reference `CharFields` or `TextFields` being covered as well

(*I'm not _entirely_ sure anymore if I tested this in the end, but I feel like I did...)

koeaw (approved):

Issue #949, opened 2024-06-12
search: related property


gythaogg:
At the moment related property field shows [all possible forward relationships](https://github.com/acdh-oeaw/apis-core-rdf/blob/main/apis_core/apis_entities/filtersets.py#L45)

Instead it should only show a subset of relations where the model can appear as subject or object .

For example, in Place search I should only see those relations where Place can appear as either subject, object or both. I should not see "author of" relationship that is defined as between Person (subj_type) and Work (obj_type).

PR #959, opened 2024-06-13
fix(apis_entities): only list possible properties in filterset form

b1rger:
The filter `related_properties` listed all the existing properties, but
it makes more sense to only list the properties that can have the model
the filter filters as a `subj` or an `obj`.

Closes: #949

### DIFF
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -1,11 +1,13 @@
 import django_filters
 from django.db import models
-from django.db.models import Q, Case, When
+from django.db.models import Q, Case, When, Value
+from django.db.models.functions import Concat
 from apis_core.generic.filtersets import GenericFilterSet, GenericFilterSetForm
 from apis_core.apis_relations.models import Property, Triple
 from apis_core.generic.helpers import generate_search_filter
 from apis_core.apis_entities.utils import get_entity_classes
 from apis_core.apis_metainfo.models import RootObject
+from django.contrib.contenttypes.models import ContentType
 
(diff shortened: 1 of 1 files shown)


b1rger:
> @b1rger In TibSchol I was asked to list the property either as forward or reverse as it applies to the entity type we are searching on. For example the property composed by/author of between Work and Person The related property field for Works will contain "composed by" and the related property field for Persons with contain "author of"
> 
> If that makes sense for APIS in general then I'd recommend that as part of #949
> 
> When the subject and object are the same type, I displayed the text to show both forward and reverse text to select the property - but in Tibschol they want to see it as two different properties to filter the search better. IMO this is probably just a Tibschol specific requirement - need not be the default behaviour of APIS core.

It makes sense to display the property with the name based on the entity we are filtering. I've tried to implement that and also added the possible property "targets". I'm not sure if it works as expected, the naming of properties in my test setups is a bit confusing

b1rger:
(pls. rereview)

b1rger:
> When the Subject and Object are the same type of entity but have different forward and reverse property labels, such as this _Character to Character_ relationship _mother of/daughter of_, I see **daughter of | character** - which is the `name_reverse`. It's all right (`name_forward` would have also been fine), just checking if this the intended behaviour.

Thats up for discussion. I've now changed that to show `name_forward / name_reverse | SubjObjType`.

> This might be because something is wrong with my setup and nothing to do with the pull request - but,
> 
>     1. when I submit without selecting any property, I get
>        > Property matching query does not exist.
> 
>     2. when I select a property and then search, I get
>        > PropertyFilter.filter() got an unexpected keyword argument 'triple_set_from_subj__prop'

I've fixed that.

PR #963, opened 2024-06-14
perf(apis_entities): refactor and cache the prev and next url methods

b1rger:


sennierer (commented):
its much more performant than the solution before, but always returns you to the detail page, instead of staying on eg the edit page (or the history).
I changed it a bit and added a templatetag that just replaces the ID of the object in the current url with the previous or next. Branch is [here](https://github.com/acdh-oeaw/apis-core-rdf/tree/ms/improve-prev-next-adapted). Doesnt even need the content type like that anymore.

b1rger:
I've integrated some of your changes, but solved it without a templatetag

sennierer (approved):
Nice! Much better without the templatetag.

Issue #676, opened 2024-02-28
django-tables2 "LinkColumn" is deprecated


koeaw:
I've been seeing Django Tables2's `LinkColumn` around in our code, which is being phased out/[deprecated](https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#linkcolumn):

> This column should not be used anymore, the `linkify` keyword argument to regular columns can be used to achieve the same results.

Just wanted to make a note of it in case it's still there.

(I'm still on version `v0.13.1` and haven't checked beyond that.)

b1rger:
```
grep -r LinkColumn *
apis_core/apis_metainfo/tables.py:    id = tables.LinkColumn()
apis_core/apis_relations/tables.py:                self.base_columns["other_entity"] = tables.LinkColumn(
apis_core/apis_relations/tables.py:                self.base_columns["other_entity"] = tables.LinkColumn(
```

b1rger:
But its still part of django-tables2 and there is nothing broken so far in our codebase, so I'm removing the `bug` label

koeaw:
True!

koeaw:
I'm leaving a snippet for `Column` here, for reference:

```py
    my_field = tables.Column(
        orderable=True,
        linkify={
            "viewname": "apis:generic:update",
            "args": [A("self_contenttype"), A("pk")],
        },
        empty_values=[None],
    )
```
I've previously used this as replacement (but am currently looking into `TemplateColumn` instead).

PR #991, opened 2024-06-20
fix(apis_relations): replace LinkColumn with Column in TripleTables

b1rger:
The LinkColumn has been deprecated by the django-tables2 project, we
should therefore replace it with Column and the `linkify` keyword. This
also lets us use the `get_absolute_url` resp. `get_edit_url` methods of
the entities.

Closes: #676


koeaw (approved):
Tested – works fine and looks just like before, which I think is exactly how it should be.

Issue #335, opened 2023-10-11
create a filter for objects changed after yyyy-mm-dd


sennierer:
after #242 is fixed a filter should be added to the generic API to filter for changes after a certain date
given that reversions (kind of) work in vanilla this should be easily possible in vanilla
related to [apis-core#404](https://github.com/acdh-oeaw/apis-core/issues/404)

PR #999, opened 2024-06-24
feat(apis_entities): add `changed since` filter to entity filterset

b1rger:
Closes: #335

### DIFF
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -2,12 +2,15 @@ import django_filters
 from django.db import models
 from django.db.models import Q, Case, When, Value
 from django.db.models.functions import Concat
+from django.conf import settings
+from django.forms import DateInput
 from apis_core.generic.filtersets import GenericFilterSet, GenericFilterSetForm
 from apis_core.apis_relations.models import Property, Triple
 from apis_core.generic.helpers import generate_search_filter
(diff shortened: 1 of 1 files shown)


apis_core/apis_entities/filtersets.py
```diff
@@ -53,6 +56,12 @@ def related_entity(queryset, name, value):
     return queryset.filter(pk__in=t)
 
 
+def changed_since(queryset, name, value):
+    history = get_history_manager_for_model(queryset.model)
+    ids = history.filter(history_date__gt=value).values_list("pk", flat=True)
+    return queryset.filter(pk__in=ids)
```
sennierer, line 62: This needs to be changed to:

```suggestion
    ids = history.filter(history_date__gt=value).values_list("id", flat=True)
```

`pk` is the primary key of the version, `id` is the primary key of the db instance

sennierer (changes requested):

apis_core/apis_entities/filtersets.py
sennierer: sorry my IDE again changed the lines.
line 61 needs to be changed to:

```python
ids = history.filter(history_date__gt=value).values_list("id", flat=True)
```

b1rger: fixed

sennierer (approved):
working as intended

PR #1009, opened 2024-06-25
make the additional columns field more flexible and adapting to model changes

### LINKED ISSUE #735: change `Columns` field in list filters to also allow for deselecting columns

sennierer:
resolves #735 
additionally allows to select annotated fields (though only tested on simple fields, no Json, Array etc)

supersedes #755

### DIFF
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -7,9 +7,12 @@ from apis_core.generic.helpers import generate_search_filter
 from apis_core.apis_entities.utils import get_entity_classes
 from apis_core.apis_metainfo.models import RootObject
 
-ABSTRACT_ENTITY_FILTERS_EXCLUDE = [
+ABSTRACT_ENTITY_COLUMNS_EXCLUDE = [
     "rootobject_ptr",
     "self_contenttype",
+]
+
+ABSTRACT_ENTITY_FILTERS_EXCLUDE = ABSTRACT_ENTITY_COLUMNS_EXCLUDE + [
     "review",
     "start_date",
     "start_start_date",
@@ -54,7 +57,7 @@ def related_entity(queryset, name, value):
 
 
 class AbstractEntityFilterSetForm(GenericFilterSetForm):
-    columns_exclude = ABSTRACT_ENTITY_FILTERS_EXCLUDE
+    columns_exclude = ABSTRACT_ENTITY_COLUMNS_EXCLUDE
 
 
 class AbstractEntityFilterSet(GenericFilterSet):
--- a/apis_core/generic/filtersets.py
+++ b/apis_core/generic/filtersets.py
@@ -21,10 +21,17 @@ class GenericFilterSet(FilterSet):
             Form = self.get_form_class()
             if self.is_bound:
                 self._form = Form(
-                    self.data, prefix=self.form_prefix, model=self._meta.model
+                    self.data,
+                    prefix=self.form_prefix,
+                    model=self._meta.model,
+                    annotations=self.queryset.query.annotations,
                 )
             else:
-                self._form = Form(prefix=self.form_prefix, model=self._meta.model)
+                self._form = Form(
+                    prefix=self.form_prefix,
+                    model=self._meta.model,
+                    annotations=self.queryset.query.annotations,
+                )
         return self._form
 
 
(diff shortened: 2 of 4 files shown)


apis_core/generic/views.py
```diff
@@ -213,7 +225,8 @@ class Autocomplete(
 
     permission_action_required = "view"
     template_name_suffix = "_autocomplete_result"
-    create_field = "thisisnotimportant"  # because we are using create_object_from_uri
```
b1rger, line 216: Please undo this change

sennierer: but thats just formatting (probably done by ruff?)
or is it that you'll prefer formatting in a different PR/commit?

sennierer: never mind, given that its just formatting removed it

apis_core/generic/filtersets.py
b1rger, line 1: The `queryset` object also contains the `model`, so it would probably make more sense to only pass the queryset to the `Form` and let the form get the `model` and the `annotations` through  the `queryset` attributes.
(Please also adapt the docstring of the `GenericFilterSet`

apis_core/generic/forms/__init__.py
```diff
@@ -40,12 +42,23 @@ class GenericFilterSetForm(forms.Form):
...
         model = kwargs.pop("model")
+        annotations = kwargs.pop("annotations", {})
         super().__init__(*args, **kwargs)
         self.fields["columns"].choices = [
             (field.name, field.verbose_name)
             for field in model._meta.fields
             if field.name not in self.columns_exclude
+        ] + [
+            (key, key) for key in annotations.keys() if key not in self.columns_exclude
         ]
```
b1rger, line 53: For readability, I would split that up:
```
modelfields = [(field.name, field.verbose_name) for field in queryset.model._meta.fields]
annotationfields = [(key, key) for key in queryset.query.annotations.keys()]
self.fields["columns"].choices = [field for field in modelfields + annotationfields if field[0] not in self.columns_exclude]
```

apis_core/generic/forms/__init__.py
```diff
@@ -40,12 +42,23 @@ class GenericFilterSetForm(forms.Form):
...
+        if not self.is_bound:
+            table_modules = module_paths(model, path="tables", suffix="Table")
+            table_class = first_member_match(table_modules, GenericTable)
+            self.fields["columns"].initial = [
+                field
+                for field in table_class.Meta.fields
+                if field not in self.columns_exclude
+            ]
```
b1rger, line 61: This part makes me unhappy, but I also don't know any other way to access the table fields. The main downside is, that is only gets a subset of fields - I can add fields in the `__init__` of the table, I can hide fields in the table before it is being rendered ...

apis_core/generic/views.py
```diff
@@ -103,13 +103,25 @@ def get_table_class(self):
...
-        columns = self.request.GET.getlist("columns", [])
+        columns = self.request.GET.getlist(
+            "columns", self.filterset.form.fields["columns"].initial
+        )
         column_fields = [
             field for field in self.model._meta.fields if field.name in columns
         ]
+        for key, val in self.object_list.query.annotations.items():
+            if key in columns:
+                field_pre = val.field if hasattr(val, "field") else val.output_field
```
b1rger, line 114: I think this would be more readable:
```
field_pre = getattr(val, "field", val.output_field)
```

b1rger (changes requested):
I begin to question my approach to put the `columns` stuff in the filterset form. Maybe it might make more sense to inject it in the view ... (but lets leave it like that for now)

apis_core/apis_entities/filtersets.py
```diff
@@ -7,9 +7,12 @@
 from apis_core.apis_entities.utils import get_entity_classes
 from apis_core.apis_metainfo.models import RootObject
 
-ABSTRACT_ENTITY_FILTERS_EXCLUDE = [
+ABSTRACT_ENTITY_COLUMNS_EXCLUDE = [
     "rootobject_ptr",
     "self_contenttype",
+]
+
+ABSTRACT_ENTITY_FILTERS_EXCLUDE = ABSTRACT_ENTITY_COLUMNS_EXCLUDE + [
```
b1rger, line 15:
```suggestion
ABSTRACT_ENTITY_FILTERS_EXCLUDE = ABSTRACT_ENTITY_COLUMNS_EXCLUDE + [ "foo",
```

b1rger (changes requested):

b1rger:
Superseded by #1014

PR #1014, opened 2024-06-27
Refactor and enhance column selector

b1rger:
- **refactor(generic): move the column selector logic to the view**
- **feat(generic): fix and extend the columns selector**
- **fix(apis_entities): split excluded filters and excluded columns**

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -1,4 +1,5 @@
 from django import http
+from django import forms
 from django.conf import settings
 from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.views.generic import DetailView
@@ -120,6 +121,26 @@ class List(
         )
         filterset_class = first_member_match(filterset_modules, GenericFilterSet)
         return filterset_factory(self.model, filterset_class)
+
+    def get_filterset(self, filterset_class):
+        """
+        We override the `get_filterset` method, so we can inject a
+        `columns` selector into the form
+        """
+        filterset = super().get_filterset(filterset_class)
+
+        # we inject a `columns` selector in the beginning of the form
+        columns = forms.MultipleChoiceField(
+            required=False,
+            choices=[
+                (field.name, field.verbose_name)
+                for field in self.model._meta.fields
+                if field.name not in filterset.form.columns_exclude
+            ]
+        )
+        filterset.form.fields = {**{"columns": columns}, **filterset.form.fields}
+
+        return filterset
 
     def get_queryset(self):
         queryset_methods = module_paths(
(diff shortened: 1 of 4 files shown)


apis_core/generic/views.py
```diff
@@ -104,14 +105,35 @@ def get_table_class(self):
...
+        )
+        modelfields = self.model._meta.fields
+        kwargs["exclude"] = [
+            field.name for field in modelfields if field.name not in selected_columns
         ]
+
+        # now we look at the selected columns and
+        # add all modelfields and annotated fields that
+        # are part of the selected columns to the extra_columns
+        annotationfields = tuple()
```
sennierer, line 123:
```suggestion
        annotationfields = list()
```
tuple is immutable, no append

apis_core/generic/views.py
```diff
@@ -104,14 +105,35 @@ def get_table_class(self):
...
+        for key, value in self.object_list.query.annotations.items():
+            fake_field = getattr(value, "field", value.output_field)
+            setattr(fake_field, "name", key)
+            annotationfields.append(fake_field)
+        extra_fields = list(
+            filter(lambda x: x.name in selected_columns, modelfields + annotationfields)
```
sennierer, line 129:
```suggestion
            filter(lambda x: x.name in selected_columns, modelfields + tuple(annotationfields))
```
as we used list to append earlier, the list needs not be converted to a tuple for combining it with modelfields

sennierer (changes requested):
when there is a annotated field this fails as tuples have no `append()`
suggested to use list and convert it back to tuple

apis_core/generic/views.py
```diff
@@ -104,14 +105,38 @@ def get_table_class(self):
...
-        column_fields = [
-            field for field in self.model._meta.fields if field.name in columns
+
+        # we look at the selected columns and exclude
+        # all modelfields that are not part of that list
+        selected_columns = self.request.GET.getlist(
+            "columns",
+            self.get_filterset(self.get_filterset_class()).form["columns"].initial,
+        )
+        modelfields = self.model._meta.fields
```
sennierer, line 115: The current solution doesnt allow to deselect all columns. If something in the form is changed the whole form is put to the GET params. So to see if we need to use the initial values or the empty columns list we check whether the GET params are empty or not.

```suggestion
        selected_columns = self.request.GET.getlist("columns") if len(
            self.request.GET) > 0 else self.get_filterset(self.get_filterset_class()).form["columns"].initial
```

sennierer (changes requested):
current solution doesnt allow to deselct all columns. See suggestions above.

sennierer (approved):

PR #1019, opened 2024-07-01
feat(sample_project): add INTERNAL_IPS to the sample_project settings

b1rger:


b1rger:
does not work wit our deployment setup

Issue #1020, opened 2024-07-01
"Additional columns" dropdown is missing many-to-many relationships


koeaw:
`ManyToManyField`s don't show up in the "additional columns" dropdown.

The `self.model._meta.fields` in this line:
https://github.com/acdh-oeaw/apis-core-rdf/blob/d0bffdb7b31ded5ab0bbfa5a498f28ec0dc9ba48/apis_core/generic/views.py#L153

only seems to return fields on the model itself.

`self.model._meta.get_fields()`, on the other hand, also includes `ManyToManyField`s but can't deal with `field.verbose_name`.

PR #1021, opened 2024-07-01
fix(generic): include many-to-many relationships in list view

koeaw:
Include ManyToMany fields in 'additional columns' dropdown as well as in table.

Closes: #1020

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -113,7 +113,7 @@ class List(
             if self.request.GET
             else self.get_filterset(self.get_filterset_class()).form["columns"].initial
         )
-        modelfields = self.model._meta.fields
+        modelfields = self.model._meta.get_fields()
         kwargs["exclude"] = [
             field.name for field in modelfields if field.name not in selected_columns
         ]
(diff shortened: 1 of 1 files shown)


koeaw:
This is a minimal fix which makes `ManyToManyField`s show up in both the dropdown and the table – if un-linkified (not sure if there is a way to linkify `ManyToManyColumn`s by default).

b1rger (changes requested):
The fix is not only for many-to-many relationships, but for all types of relationships. Can you change the commit message to reflect that?

Ideally the commit message would explain why you also had to change the tuple in the list comprehension to use `getattr`

koeaw:
Can do!

koeaw:
> The fix is not only for many-to-many relationships, but for all types of relationships. 

I've adapted the commit message, but I don't know what exactly this means because the Django source code only mentions `ManyToManyFields` in [`fields`](https://github.com/django/django/blob/aa74c4083e047473ac385753e047e075e8f04890/django/db/models/options.py#L517-L525).

I get the error for `verbose_names` I got was for `ManyToOneRel`, i.e. FK, but don't quite understand how that comes about, tbh.

b1rger (approved):

b1rger:
> > The fix is not only for many-to-many relationships, but for all types of relationships.
> 
> I've adapted the commit message, but I don't know what exactly this means because the Django source code only mentions `ManyToManyFields` in [`fields`](https://github.com/django/django/blob/aa74c4083e047473ac385753e047e075e8f04890/django/db/models/options.py#L517-L525).
> 
> I get the error for `verbose_names` I got was for `ManyToOneRel`, i.e. FK, but don't quite understand how that comes about, tbh.

Ah, I thought there were other field types that are m2m, my mistake.

Issue #1022, opened 2024-07-01
Column sorting seems broken


koeaw:
I just tried sorting my table columns on `v0.23.0` by clicking on the column headers, which made all columns except for the icon columns disappear and emptied the "additional columns" dropdown.

Tried this with several tables and the behaviour seems to slightly differ between them, e.g. for ones with custom columns all columns except those custom ones (and the icon columns) disappear.

I.e. seems like columns for "proper" fields are affected? Will investigate further.

b1rger:
Ah, yeah. This is due to this assignment:
https://github.com/acdh-oeaw/apis-core-rdf/blob/d0bffdb7b31ded5ab0bbfa5a498f28ec0dc9ba48/apis_core/generic/views.py#L111

if there is a get parameter, but it does not contain the `columns`, that is interpreted as an emtpy list of columns. I guess we should find out how to append the sorting of the table to the existing parameters

koeaw:
Hmm, so this seems to happen when I try to sort on the default view of a table.

When I first actively add or remove a column via the "addional columns" dropdown, and only _then_ sort, all columns in the table remain in place.

PR #1027, opened 2024-07-02
fix(generic): repair table columns ordering

b1rger:
We interpreted a GET request *with* parameters but *without* a columns
parameter as the wish to have a table without any columns. But that
could also be a wish to order the table, which only appends a `sort`
parameter to existing parameters (or creates parameters if there are
none).
Therefore for now we fallback to the default columns if none are
selected, but we gain table ordering.

Closes: #1022

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -108,10 +108,9 @@ class List(
 
         # we look at the selected columns and exclude
         # all modelfields that are not part of that list
-        selected_columns = (
-            self.request.GET.getlist("columns")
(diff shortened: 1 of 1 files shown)


koeaw (approved):
This works for me! Fixes both column sorting and paging, which I only just noticed was broken too.

I think an additional improvement could be to keep the currently included columns checked in the "additional columns" dropdown – they become unchecked on sorting or paging. But that's more a nice-to-have, so could go in a separate PR.

PR #1054, opened 2024-07-25
chore: include all contributors in the pyproject.toml

sennierer:


koeaw:
I don't use my main email address for GitHub/dev stuff, could you please swap it out (looking up the other one, just a sec) + would prefer to have only my initial used for first name.

koeaw:
> looking up the other one, just a sec

**dev-kk @** is what I use @sennierer

sennierer:
changed it

koeaw:
> changed it

Thx! For name I only meant the first name initial, but I can adapt that another time, email was the more critical thing.

b1rger (changes requested):
if we are at it, we could fix the field to [adhere to the spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#authors-maintainers)
> These keys accept an array of tables with 2 keys: name and email. Both values must be strings. The name value MUST be a valid email name (i.e. whatever can be put as a name, before an email, in [RFC 822](https://datatracker.ietf.org/doc/html/rfc822.html)) and not contain commas. The email value MUST be a valid email address. Both keys are optional, but at least one of the keys must be specified in the table.

```
authors = [
{name = "Matthias Schlögl",  email = "matthias.schloegl@oeaw.ac.at"}, 
{name = "Birger Schacht",  email = "birger.schacht@oeaw.ac.at"}, 
{name = "KK", email = "dev-kk@oeaw.ac.at"},
{name = "Saranya Balasubramanian", email ="saranya.balasubramanian@oeaw.ac.at"}
]
```

sennierer:
but in the spec you referenced it says under point 3:
`If both email and name are provided, the value goes in [Author-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-author-email) or [Maintainer-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-maintainer-email) as appropriate, with the format {name} <{email}>.`

b1rger:
> but in the spec you referenced it says under point 3: `If both email and name are provided, the value goes in [Author-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-author-email) or [Maintainer-email](https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-maintainer-email) as appropriate, with the format {name} <{email}>.`

Yes, this describes how the packaging tools use that data and how they put it in the package metadata

b1rger:
As @sennierer pointed out, I was wrong, poetry does not adhere to the standard, we have to use the string list :(

b1rger (approved):

PR #1061, opened 2024-07-30
feat(apis_entites): add a generic entites autocomplete endpoint

b1rger:
This endpoint allows us to use autocomplete over multiple model types.
It takes a parameter `entities` which is a list of ContentType natural
keys and searches for the query in all instances of those entities
(using `generate_search_filter`, which means it uses a different search
approach for every model).
The return values of the endpoint are then prefixed with the id of the
contenttype of the results, separated by an underscore.

### DIFF
--- a/apis_core/apis_entities/urls.py
+++ b/apis_core/apis_entities/urls.py
@@ -10,6 +10,7 @@ from apis_core.apis_entities.views import (
     EntitiesDuplicate,
     EntitiesUpdate,
     EntitiesMerge,
+    EntitiesAutocomplete,
 )
 
 
@@ -88,4 +89,5 @@ urlpatterns = [
         "entity/<entitytocontenttype:contenttype>/",
         include(entity_patterns),
     ),
+    path("autocomplete/", EntitiesAutocomplete.as_view(), name="autocomplete"),
 ]
--- a/apis_core/apis_entities/views.py
+++ b/apis_core/apis_entities/views.py
@@ -1,13 +1,18 @@
 from django.contrib import messages
 from django.contrib.auth.mixins import PermissionRequiredMixin
+from django.contrib.contenttypes.models import ContentType
+from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
 from django.utils.html import format_html
 from django.views import View
 from django.views.generic.edit import FormView
+from dal import autocomplete
 
 from apis_core.generic.views import GenericModelMixin, Update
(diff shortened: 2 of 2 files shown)


apis_core/apis_entities/views.py
```diff
@@ -73,3 +78,41 @@ def get_success_url(self):
...
+    Example:
+    Using this endpoint with the parameters:
+
+        ?entities=apis_ontology.person&entities=apis_ontology.place&q=ammer
+
+    gives you all the persons and places that have `ammer` in their names
+    and labels.
+    """
+
+    def get_result_value(self, result):
```
sennierer, line 102:
```suggestion
    def get_result_value(self, result) -> str:
```

apis_core/apis_entities/views.py
```diff
@@ -73,3 +78,41 @@ def get_success_url(self):
...
+        content_type = ContentType.objects.get_for_model(result)
+        return f"{content_type.id}_" + super().get_result_value(result)
+
+    def get_queryset(self):
```
sennierer, line 106:
```suggestion
    def get_queryset(self) -> RootObject:
```

apis_core/apis_entities/views.py
```diff
@@ -73,3 +78,41 @@ def get_success_url(self):
             "apis_core:apis_entities:generic_entities_edit_view",
             args=[self.get_object().__class__.__name__.lower(), self.get_object().id],
         )
+
+
+class EntitiesAutocomplete(autocomplete.Select2QuerySetView):
+    """
+    This endpoint allows us to use autocomplete over multiple model types.
```
sennierer, line 85:
```suggestion
    This endpoint allows us to use autocomplete over multiple model classes.
```

apis_core/apis_entities/urls.py
```diff
@@ -88,4 +89,5 @@ def to_url(self, value):
         "entity/<entitytocontenttype:contenttype>/",
         include(entity_patterns),
     ),
+    path("autocomplete", EntitiesAutocomplete.as_view(), name="autocomplete"),
```
sennierer, line 92:
```suggestion
    path("autocomplete/", EntitiesAutocomplete.as_view(), name="autocomplete"),
```

not working in my current setup without the trailing slash. Though it worked before?
However, trailing slashes in all other routes, so should be there anyway

sennierer (changes requested):
not working in my case without the trailing slash in the URLs. Worked before when I tested the new relations. However, for consistency, please add the trailing slash.

sennierer (approved):

PR #1062, opened 2024-07-31
feat(relations): introduce new relations implementation

b1rger:


apis_core/apis_entities/templates/apis_core/apis_entities/abstractentity_form.html
```diff
@@ -11,6 +11,7 @@
 {% endblock col-zero %}
 
 {% block col-one %}
+  {% include "relations/list_relations_include.html" with edit=True %}
```
sennierer, line 14: do we the possibility to not use the new relations? or do we switch and need to use them as soon as this PR is merged? if the former this needs to be changed as it will break is relations is not in installed apps

apis_core/apis_entities/templates/apis_core/apis_entities/abstractentity_detail.html
```diff
@@ -10,6 +10,7 @@
 {% endblock col-zero %}
 
 {% block col-one %}
+  {% include "relations/list_relations_include.html" %}
```
sennierer, line 13: do we the possibility to not use the new relations? or do we switch and need to use them as soon as this PR is merged? if the former this needs to be changed as it will break is relations is not in installed apps

apis_core/relations/forms.py
```diff
@@ -0,0 +1,182 @@
...
+    validate (some validation happens in the `clean` method anyway).
+    """
+
+    def validate(self, value):
+        return True
+
+
+class RelationForm(forms.ModelForm):
+    """
+    This form overrides generic form for relation editing.
```
sennierer, line 26: Is this correct? It inherits from ModelForm, not from generic form

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+                )
+        if self.obj_content_type:
+            if self.obj_content_type.model_class() not in self.obj_list():
+                raise ValidationError(
+                    f"{self.obj} is not of any type in {self.obj_list()}"
+                )
+        super().save(*args, **kwargs)
+
+    @property
+    def subj_to_obj_text(self):
```
sennierer, line 60:
```suggestion
    def subj_to_obj_text(self) -> str:
```

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+        if hasattr(self, "name"):
+            return f"{self.subj} {self.name()} {self.obj}"
+        return f"{self.subj} relation to {self.obj}"
+
+    @property
+    def obj_to_subj_text(self):
```
sennierer, line 66:
```suggestion
    def obj_to_subj_text(self) -> str:
```

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+        if hasattr(self, "reverse_name"):
+            return f"{self.obj} {self.reverse_name()} {self.subj}"
+        return f"{self.obj} relation to {self.subj}"
+
+    def __str__(self):
+        return self.subj_to_obj_text
+
+    @classmethod
+    def subj_list(cls):
```
sennierer, line 75:
```suggestion
    def subj_list(cls) -> list[models.Model]:
```

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+        return cls.subj_model if isinstance(cls.subj_model, list) else [cls.subj_model]
+
+    @classmethod
+    def obj_list(cls):
```
sennierer, line 79:
```suggestion
    def obj_list(cls) -> list[models.Model]:
```

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+        return cls.obj_model if isinstance(cls.obj_model, list) else [cls.obj_model]
+
+    @classmethod
+    def name(cls):
```
sennierer, line 83:
```suggestion
    def name(cls) -> str:
```

apis_core/relations/models.py
```diff
@@ -0,0 +1,88 @@
...
+        return cls._meta.verbose_name
+
+    @classmethod
+    def reverse_name(cls):
```
sennierer, line 87:
```suggestion
    def reverse_name(cls) -> str:
```

apis_core/relations/tables.py
```diff
@@ -0,0 +1,12 @@
+from apis_core.generic.tables import GenericTable
+
+
+class RelationsListTable(GenericTable):
+    class Meta:
+        attrs = {"class": "table-sm"}
+        sequence = GenericTable.Meta.sequence
+
+    def render_desc(self, record):
```
sennierer, line 9:
```suggestion
    def render_desc(self, record) -> str:
```

apis_core/relations/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        issubclass(mc, Relation)
+        and hasattr(mc, "subj_model")
+        and hasattr(mc, "obj_model")
+    )
+
+
+@functools.cache
+def relation_content_types(
+    subj_model=None, obj_model=None, any_model=None, combination=(None, None)
+):
```
sennierer, line 18:
```suggestion
) -> set[ContentType]:
```

apis_core/relations/views.py
```diff
@@ -0,0 +1,73 @@
...
+
+from apis_core.relations.templatetags.relations import (
+    relations_from,
+    possible_relation_types_from,
+)
+from apis_core.relations.forms import RelationFormHX
+
+
+class CreateRelation(Create):
+    def get_form_kwargs(self, *args, **kwargs):
```
sennierer, line 15:
```suggestion
    def get_form_kwargs(self, *args, **kwargs) -> dict:
```

sennierer (commented):
works really nice!

needed to additionally use #1061 to make it work (autocompletes from that PR are expected by this one)

tested this with:
- VersionMixin and without, both worked. VersionMixin created correct versions
- with an additional date field in the relation, rendered fine in the form, saved correct
- with a list of models in the object position, worked as expected.

slight improvements that can also be addressed with subsequent PRs:
- the default table of the relations could show the clickable subj/obj of the relation and a name/reverse_name instead of just printing the textual representation
- when the PR is merged we need to adapt the history module to include the new relations
- there should be links to the list views of the relations (the generic ones) similar to the entities from the header of the webpage
- management function that creates Relation classes from existing triples (as code)
- management function that copies triples to new relations

added minor suggestions in the code (mainly some type hints), that can also be omitted,

sennierer:
created some issues to fix either in this PR or subsequent ones:
- [ ] #1084 
- [ ] #1083 
- [ ] #1082 
- [ ] #1081 
- [ ] #1080 
- [ ] #1079

sennierer (changes requested):
please add the suggested fix to the select2 function to re-enable the text searching in the dropdown

apis_core/relations/templates/apis_core/apis_entities/abstractentity_form.html
```diff
@@ -0,0 +1,30 @@
...
+  {{ block.super }}
+{% endblock col-one %}
+
+{% block scripts %}
+  <script>
+  document.body.addEventListener("dismissModal", function(evt){
+    console.log("triggered dismissModal");
+    $('#modal').modal('hide');
+  });
+  function reinitSelect2(something) {
```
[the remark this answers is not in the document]
b1rger: Yes, we now send the form id in the trigger and only reinitialize the selects in this form

b1rger:
I've now also implemented the possibility to override the RelationsListTable. A table from a person to events should be called `PersonEventRelationsTable`

sennierer (approved):
This works nicely. Only minor issue is that the select2 form field renders still as two fields when the form doesnt validate. It works, but doesnt look nice.
I tracked this down to the fact that the `select2-hidden-accessible` class on the select element is missing. This class secures that only the child span field is rendered and not the parent select. Tried to add the class again using an event listener on the htmx ajax event, but this didnt work.
I suggest to finalize this PR with the slight problem and create an issue for that.

docs/source/ontology.rst
```diff
@@ -40,3 +40,38 @@ updating, merging and deleting entities.
...
+   from apis_core.relations.models import Relation
+
+   class PersonLivedInPlace(Relation):
+        subj_model = Person
+        obj_model = Place
+
+You can define the class methods `name` and `reverse_name` to provide human readable
+strings for your relation model. They defaulto to the `verbose_name` (`name`) and the
+`verbose_name` with the string ` reverse` appended (`reverse_name`).
+
```
sennierer, line 71:
```suggestion

.. code-block:: python

   from apis_core.relations.models import Relation

   class PersonLivedInPlace(Relation):
        subj_model = Person
        obj_model = Place

        @classmethod
        def name(self) -> str:
            return "lived in"

        @classmethod
        def reverse_name(self) -> str:
            return "had inhabitant"

```

docs/source/ontology.rst
```diff
@@ -40,3 +40,38 @@ updating, merging and deleting entities.
...
+.. code-block:: python
+
+   from apis_core.relations.models import Relation
+
+   class PersonLivedInPlace(Relation):
+        subj_model = Person
+        obj_model = Place
+
+You can define the class methods `name` and `reverse_name` to provide human readable
+strings for your relation model. They defaulto to the `verbose_name` (`name`) and the
```
sennierer, line 69:
```suggestion
strings for your relation model. They defaul to to the `verbose_name` (`name`) and the
```

sennierer (approved):

PR #1073, opened 2024-08-12
docs: move history and collections features to dedicated page

sennierer:
features that are shipped with APIS, but need to be activated and/or configured are moved to a dedicated "Additional Features" page


docs/source/features.rst
```diff
@@ -0,0 +1,114 @@
...
+===================
+
+APIS is designed to be extensible and customizable. This section describes some 
+additional features that are shipped with APIS and how to use them.
+
+History plugin
+--------------
+
+The :py:mod:`apis_core.history` module provides versioning 
+to the APIS framework. It is based on the `django-simple-history`_
```
b1rger, line 11:
```suggestion
for APIS. It is based on the `django-simple-history`_
```

docs/source/features.rst
```diff
@@ -0,0 +1,114 @@
...
+
+
+
+API endpoint
+^^^^^^^^^^^^
+
+The :class:`apis_core.history.api_views.GenericHistoryLog` class is a viewset
+that provides a REST API endpoint for the version history of a model. It can be
+used to retrieve the version history of a model instance.
+The viewset can be accessed under `/apis/api/history/entity/edit_log/`. It takes
```
b1rger, line 40: I don't find `edit_log` anywhere in the codebase, I guess that has changed?

b1rger (changes requested):

docs/source/features.rst
sennierer: good catch, changed the url to the one that is actually used

b1rger (changes requested):
This is now missing the `features.rst` file

sennierer:
> This is now missing the `features.rst` file

added it now

b1rger (changes requested):
I guess we can remove the `history.rst` and the `collections.rst` files in the same commit? The `features.rst` is basically a combination of those two, isn't it?

sennierer:
yes, I will remove the files

b1rger (approved):

Issue #1066, opened 2024-08-12
Add possibility to tag all added entities/relations in a session


b1rger:
That should be implemented in the `collections` module using signals and a simple interface

PR #1074, opened 2024-08-12
feat(collections): allow add all changes in a session to a collection

b1rger:
This commit implements a way to add SkosCollections to the users
session. If there are SkosCollections in the session and an model that
inherits from `apis_core.history.VersionMixin` get saved, the created
version of that object gets tagged with the SkosCollections.

Closes: https://github.com/acdh-oeaw/apis-core-rdf/issues/1066

### DIFF
--- a/apis_core/collections/signals.py
+++ b/apis_core/collections/signals.py
@@ -1,0 +1,30 @@
+from django.contrib import messages
+from django.contrib.contenttypes.models import ContentType
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from crum import get_current_request
+from apis_core.collections.models import SkosCollection, SkosCollectionContentObject
+from apis_core.history.models import APISHistoryTableBase
+
+
+@receiver(post_save)
+def add_to_session_collection(
+    sender, instance, created, raw, using, update_fields, **kwargs
+):
+    """
+    Add a created apis_core.history model instance to all the SkosCollections
+    that are listed in the `session_collections` session variable.
+    This needs the 'crum.CurrentRequestUserMiddleware' middleware to
+    be enabled.
+    """
+    request = get_current_request()
+    if isinstance(instance, APISHistoryTableBase) and request:
+        for pk in request.session.get("session_collections", []):
+            sc = SkosCollection.objects.get(pk=pk)
+            content_type = ContentType.objects.get_for_model(instance)
+            SkosCollectionContentObject.objects.create(
+                collection=sc,
+                content_type=content_type,
+                object_id=instance.history_id,
+            )
+            messages.info(request, f"Tagged {instance} with tag {sc}")
--- a/apis_core/collections/templatetags/apis_collections.py
+++ b/apis_core/collections/templatetags/apis_collections.py
@@ -93,3 +93,17 @@ def collection_content_objects(context, obj, collectionids=None):
         ids = collectionids.split(",")
         sccos = sccos.filter(collection__id__in=ids)
     return sccos
+
+
+@register.inclusion_tag(
+    "collections/collection_session_toggle.html", takes_context=True
+)
+def collection_session_toggle_by_id(context, collection_id):
+    """
+    Provide a checkbox to toggle if a session collection is active.
+    The checkbox calls the CollectionSessionToggle view.
+    """
+    session_collections = context.request.session.get("session_collections", [])
+    context["collection"] = SkosCollection.objects.get(pk=collection_id)
+    context["enabled"] = collection_id in session_collections
+    return context
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -112,3 +112,11 @@ implement a workflow (i.e. three collections: `done` as the root one, `in
 process` with `done` as parent and `todo` with `in process` as parent - the
 user can then on the click of a button change an the collection an instance is
 connected to)
+
+* :py:func:`apis_core.collections.templatetags.apis_collections.collection_session_toggle_by_id`
+
+This templatetag provides a checkbox that enables a collection as "session collection" - this
+means that if enabled, all new versions of instances will be added to this collection. This is
+implemented in :py:func:`apis_core.collections.signals.add_to_session_collection`. To use this
+feature, :py:mod:`apis_core.history` has to be enabled and the
+:py:class:`crum.CurrentRequestUserMiddleware` has to be added to the ``MIDDLEWARE``
--- a/apis_core/collections/apps.py
+++ b/apis_core/collections/apps.py
@@ -4,3 +4,6 @@ from django.apps import AppConfig
 class VocabsConfig(AppConfig):
     default_auto_field = "django.db.models.AutoField"
     name = "apis_core.collections"
+
+    def ready(self):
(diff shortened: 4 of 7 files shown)


b1rger:
@sennierer I need input: this PR works, but for now the signal only tags instances inheriting from `AbstractEntity`. We only talked about tagging relations, should we change it to tag child instances of `Triple`? Or should we instead hook it to the `VersionMixin` and tag every new history object (then we would catch `TempTriples` and most entities...)

sennierer:
> @sennierer I need input: this PR works, but for now the signal only tags instances inheriting from `AbstractEntity`. We only talked about tagging relations, should we change it to tag child instances of `Triple`? Or should we instead hook it to the `VersionMixin` and tag every new history object (then we would catch `TempTriples` and most entities...)

Hooking it to `VersionMixin` would actually be a nice idea, as it would allow to even distinguish between entities that have been created/edited for one project and subsequently changed for another. I guess that wouldnt be a problem if the history plugin is not activated in an instance? It would then just not work?

b1rger:
> I guess that wouldnt be a problem if the history plugin is not activated in an instance? It would then just not work?

Exactly, I'd just do an `if isinstance(instance, VersionMixin)` in the signal

b1rger:
> > I guess that wouldnt be a problem if the history plugin is not activated in an instance? It would then just not work?
> 
> Exactly, I'd just do an `if isinstance(instance, VersionMixin)` in the signal

This is now implemented. #1078 might make it easier to review

b1rger:
to test:
create an skoscollection; add `{% load apis_collections %}` and `{% collection_session_toggle_by_id <skoscollectionid> %}` to a template

sennierer (changes requested):
In my implementation the routes were not working. See the pushed changes for suggested fix: include the collections url in apis_core.url if collections module is in installed apps, add an app_name to urls in collections.urls and use the namespaced urlnames in the templates.

sennierer:
Another thing I just realized: The toggle_session button is visualizing if the collection is activated for the current session, but it is actually not showing whether the entity has actually been added to the collection or not. Given that the other collection toggle buttons look the same, but visualize whether the entity has been added to the collection this is going to be difficult to grasp for the users. I suggest the toggle_session button should visualize two state, whether the collection is activated for the current session (eg via a checkbox) and whether the current entity has already been added to the collection or not. Consequently a click on the checkbox should toggle whether its active for the session and a click on the button could toggle whether to add the entity/relation to the collection or not (this would allow to also add entities/relations that are not altered).

Given the fact that one page can show several entities/relations we can also keep a simple version of the button that just shows whether a certain collection is active for the session or not, but this should be a checkbox (or something else that is different than the normal toggle collection button)

b1rger:
Add messages that list changes added to collection

apis_core/collections/templatetags/apis_collections.py
```diff
@@ -93,3 +93,17 @@ def collection_content_objects(context, obj, collectionids=None):
...
         sccos = sccos.filter(collection__id__in=ids)
     return sccos
+
+
+@register.inclusion_tag(
+    "collections/collection_session_toggle.html", takes_context=True
+)
+def collection_session_toggle_by_id(context, collection_id):
+    """
+    Provide a button to toggle if a session collection is active.
```
sennierer, line 103:
```suggestion
    Provide a checkbox to toggle if a session collection is active.
```

sennierer (commented):
I didnt see the messages that list the changes added to the collection. How is that implemented?

sennierer (commented):
ups...sorry forget the last message, my PR plugin didnt pull latest changes

sennierer (commented):
ok, pulled the latest code, selectbox is now there, but the rest is still the case: first click on the select fires get request, but select box gets not ticked, second again fires GET request and select box gets ticked. However, any reload and selectbox is unticked again and no save of the object to the collection. Havent debugged it yet, but looks like the collection doesnt get saved in the session?

sennierer (commented):
...and removing the quotes keeps the checkbox activated, but changing the objects does not add anything to the collection.

apis_core/collections/signals.py
```diff
@@ -0,0 +1,29 @@
...
+@receiver(post_save)
+def add_to_session_collection(
+    sender, instance, created, raw, using, update_fields, **kwargs
+):
+    """
+    Add a created object to all the SkosCollections that are listed
+    in the `session_collections` session variable.
+    """
+    request = get_current_request()
+    if isinstance(instance, VersionMixin) and request:
```
sennierer, line 19: This check here never hits. The instance is of `VersionPerson` rather than `VersionMixin`

sennierer: so when changing something in a Person there is two calls to this signal, one with `isinstance(instance, Person)` and one with `isinstance(instance, VersionPerson)`

apis_core/collections/signals.py
```diff
@@ -0,0 +1,29 @@
...
+
+@receiver(post_save)
+def add_to_session_collection(
+    sender, instance, created, raw, using, update_fields, **kwargs
+):
+    """
+    Add a created object to all the SkosCollections that are listed
+    in the `session_collections` session variable.
+    """
+    request = get_current_request()
```
sennierer, line 18: ok, tracked it down.
I used the sample_project as is and in the sample project the crum middleware:
```python
MIDDLEWARE += ('crum.CurrentRequestUserMiddleware',)
```
is missing. And apparently the `get_current_request` function just returns False instead of raising an exception if thats missing.

apis_core/collections/signals.py
sennierer:
```suggestion
    if hasattr(instance, "history_id") and request:
```
would allow to further down just use the object as is (the history object just created) instead of hitting the DB for retrieving the latest history object.

sennierer (approved):
Nice!

docs/source/features.rst
```diff
@@ -111,4 +111,12 @@ a collection to point to the collections parent. This is useful if you want to
...
\ No newline at end of file
+connected to)
+
+* :py:func:`apis_core.collections.templatetags.apis_collections.collection_session_toggle_by_id`
+
+This templatetag provides a checkbox that enabled a collection as "session collection" - this
+means that if enabled, all new version of instances will be added to this collection. This is
+implemented in :py:func:`apis_core.collections.signals.add_to_session_collection`. To use this
+feature, :py:mod:`apis_core.history` has to be enabled and the
+:py:class:`crum.CurrentRequestUserMiddleware` has to be added to the ``MIDDLEWARE``
```
sennierer, line 122:
```suggestion

* :py:func:`apis_core.collections.templatetags.apis_collections.collection_session_toggle_by_id`

This templatetag provides a checkbox that enables a collection as "session collection" - this
means that if enabled, all new versions of instances will be added to this collection. This is
implemented in :py:func:`apis_core.collections.signals.add_to_session_collection`. To use this
feature, :py:mod:`apis_core.history` has to be enabled and the
:py:class:`crum.CurrentRequestUserMiddleware` has to be added to the ``MIDDLEWARE``
```

sennierer (approved):

PR #1078, opened 2024-08-13
feat(collections): introduce custom collections table

b1rger:
The custom collections table contains a `target` column that lists, with
a link, the instance the collection is pointing to.


sennierer (approved):
works as expected

PR #1085, opened 2024-08-14
chore: add spectacular settings for working swagger def to demo-project

sennierer:


b1rger (changes requested):
We don't really need all those settings, do we? Why would we want to set the Version to 0.13 for example?
I think its enough to set:
```
# for django spectacular to be able to generate the schema, we have to use its view inspector
REST_FRAMEWORK = {"DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema"}
# we use our custom schema generator to make it pick up our custom routes
SPECTACULAR_SETTINGS = {"DEFAULT_GENERATOR_CLASS": "apis_core.generic.generators.CustomSchemaGenerator"}
```

Please also add `sample_project` as the scope to the commit message and remove the `to demo-project`

sennierer:
> We don't really need all those settings, do we? Why would we want to set the Version to 0.13 for example? I think its enough to set:
> 
> ```
> # for django spectacular to be able to generate the schema, we have to use its view inspector
> REST_FRAMEWORK = {"DEFAULT_SCHEMA_CLASS": "drf_spectacular.openapi.AutoSchema"}
> # we use our custom schema generator to make it pick up our custom routes
> SPECTACULAR_SETTINGS = {"DEFAULT_GENERATOR_CLASS": "apis_core.generic.generators.CustomSchemaGenerator"}
> ```
> 
> Please also add `sample_project` as the scope to the commit message and remove the `to demo-project`

true, changed that

b1rger (approved):

PR #1097, opened 2024-08-23
Ms/simple management command for ng relation classes

sennierer:


sennierer:
Making this a draft is we most likely wont merge this into main (as it is only needed once and most probably needs adaption for the various project), but could be used as a starting point for projects to create new relation classes.

PR #1103, opened 2024-08-27
birger/sample project relations

b1rger:
- chore(sample_project): implement relations in models.py & add migrations
- chore(sample_project): update fixtures with relations instead of triples


sennierer (approved):
Working as expected. we should note down somewhere (maybe in the docs?) that for loading the fixtures the history module needs to be deactivates: `SIMPLE_HISTORY_ENABLED = False` in the settings.

PR #1110, opened 2024-08-27
docs(history): add documentation on some management commands

### LINKED ISSUE #873: Document management command for activating history module
There's a management command which needs to be run to "initialise" history when `VersionMixin` is introduced to models with existing data.

The [docs](https://acdh-oeaw.github.io/apis-core-rdf/history.html) currently don't make mention of that.

sennierer:
resolves #873

### DIFF
--- a/docs/source/features.rst
+++ b/docs/source/features.rst
@@ -48,6 +48,41 @@ can be accessed under `/apis/swagger/schema/swagger-ui/#/apis/apis_api_history_e
 .. _django-simple-history: https://django-simple-history.readthedocs.io/en/latest/
 
 
+Management Commands
+^^^^^^^^^^^^^^^^^^^
+
+The APIS history plugin, based on `django-simple-history`, provides several management commands to help curate history objects. Here are some of the most useful commands:
+
+1. Populate History Tables
+--------------------------
+
+.. code-block:: bash
+
+  python manage.py populate_history --auto
+
+This command populates history tables with the current state of the models. It's particularly useful when you've added `VersionMixin` to an existing model. The `--auto` flag applies the command to all tracked models, or you can specify a list of models instead.
+
+2. Clean Duplicate History Entries
+----------------------------------
+
+.. code-block:: bash
+
+  python manage.py clean_duplicate_history --auto
+
+This command removes duplicate history entries. `django-simple-history` creates a history object every time `save()` is called on an object, regardless of whether any changes were made. This command deletes history objects that are identical to the previous entry.
+
+3. Remove Old History Entries
+-----------------------------
+
+.. code-block:: bash
+
+  python manage.py clean_old_history --days 60 --auto
+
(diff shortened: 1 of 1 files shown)


docs/source/features.rst
```diff
@@ -48,6 +48,32 @@ can be accessed under `/apis/swagger/schema/swagger-ui/#/apis/apis_api_history_e
 .. _django-simple-history: https://django-simple-history.readthedocs.io/en/latest/
 
 
+Management Commands
+^^^^^^^^^^^^^^^^^^^
+
+The APIS history plugin is based on the `django-simple-history` package. It comes with 
+management commands provided by `django-simple-history`. Some of the more useful commands are:
```
koeaw, line 55: The first sentence repeats what's already been said at the top of the page. More helpful here would be a direct link to django-simple-history's management commands (docs) for further exploration.

Ex.
The APIS history plugin reuses `django-simple-history`'s [management commands]. Some of the more useful ones are:

docs/source/features.rst
```diff
@@ -48,6 +48,32 @@ can be accessed under `/apis/swagger/schema/swagger-ui/#/apis/apis_api_history_e
...
+  python manage.py populate_history --auto
+
+This command will populate the history tables with the current state of the models. This is useful 
+if you have added the `VersionMixin` to an existing model.
+
+.. code-block:: bash
+
+  python manage.py clean_duplicate_history --auto
+
+This command will clean up duplicate history entries.
```
koeaw, line 68: Could there be a 1-sentence or half sentence explanation about duplicates added?

Why are there duplicates/how do they come about/why do they need cleaning? Or alternatively a link to where the django-simple-history docs talk about them.

docs/source/features.rst
```diff
@@ -48,6 +48,32 @@ can be accessed under `/apis/swagger/schema/swagger-ui/#/apis/apis_api_history_e
...
+
+.. code-block:: bash
+
+  python manage.py clean_old_history --days 60 --auto
+
+This command will clean up history entries older than 60 days.
+
+
```
koeaw, line 76: All of the commands mentioned feature the argument `--auto`, but its purpose is never explained.

Would add a note either to the description of the first command (first appearance) or here at the end – or leave if off altogether (if possible/sensible).

docs/source/features.rst
b1rger: I would not use the term `reuse` for the management commands. The history app does not use those commands, they simply exist, and that only if the `simple_history` app is enabled.

docs/source/features.rst
sennierer: I dont think that there is an answer to this. Depending on your code you might call save on an object twice (eg via signals) and get duplicates. This management command just checks whether there are exactly the same objects in your history tables and removed one.

sennierer:
Ran the whole part of the documentation through an LLM to improve writing and structure.

Issue #1109, opened 2024-08-27
Double inheritance of models stops autocomplete working on child entity class


richardhadden:
Error produced when using entity autocomplete endpoint:

django.core.exceptions.FieldError: Cannot resolve keyword 'person' into field. Choices are: entity, id, property, self_contenttype, self_contenttype_id, statement, triple_set_from_obj, triple_set_from_subj, uri
[27/Aug/2024 12:11:46] "GET /entities/autocomplete/?entities=manmax_ontology.person&term=max&_type=query&q=max HTTP/1.1" 500 138119

```python
class Entity(VersionMixin, AbstractEntity):
    label = models.CharField(max_length=500, blank=False, null=False)

    def __str__(self) -> str:
        return self.label
        
class Person(Entity):
    def __str__(self) -> str:
        return self.label
```

Error seems to be in `apis_entities.views` (https://github.com/acdh-oeaw/apis-core-rdf/blob/8559c4947948e890c9694a09eeaf5110bed96b27/apis_core/apis_entities/views.py#L114)

Trying to get the correct child class (`Person` in this case) does not work

b1rger:
One problem is, that we have to filter for the model type (the `{name}__isnull`) together with the filter for the search string (which uses the `generate_search_filter` method). Because if in `ModelA`, the default is to search in `name` and in `ModelB` the default is to search in `label`, the query would break otherwise, so we have to combine `ModelA` with the search in `name` and `ModelB` with the search in `label`

richardhadden:
https://django-model-utils.readthedocs.io/en/latest/managers.html#inheritancemanager

btw, it `select_subclasses` doesn't work as we seemed to think — it doesn't _filter_ by subclasses, just returns the particular subclass-types specified as that class, rather than the generic class

b1rger:
> https://django-model-utils.readthedocs.io/en/latest/managers.html#inheritancemanager
> 
> btw, it `select_subclasses` doesn't work as we seemed to think — it doesn't _filter_ by subclasses, just returns the particular subclass-types specified as that class, rather than the generic class

ah, good to know, thats probably why there was a rootobject

richardhadden:
```python
q |= Q(**{f"self_contenttype": content_type}) & generate_search_filter(
                    content_type.model_class(), self.q, prefix=f"{name}__"
                )
```
(just filter by `self_contenttype` from `RootObject`)

[...]

PR #1111, opened 2024-08-27
fix(apis_entities): make autocomplete work with multilevel inheritance

b1rger:
Closes: #1109


sennierer (approved):
tested with a `SubPlace` that inherites from `Place` in the `sample_project`. In 8559c4947948e890c9694a09eeaf5110bed96b27 issue was reproduceable, fixed in this commit

PR #1114, opened 2024-08-28
fix(relations): make `get_relation_targets_from` method more targeted

b1rger:
The `get_relation_targets_from` method listed all possible targets of
all the relations an object might be related to. But we actually want
only a list of the targets of the relations *other* sides, seen from the
objects point of view. This is why we now also instead pass the object to
`get_relation_targets_from` and use that to check for the right targets.

### DIFF
--- a/apis_core/relations/templatetags/relations.py
+++ b/apis_core/relations/templatetags/relations.py
@@ -15,11 +15,14 @@ def possible_relation_types_from(obj) -> list[ContentType]:
 
 
 @register.simple_tag
-def get_relation_targets_from(relations) -> list[ContentType]:
+def get_relation_targets_from(obj) -> list[ContentType]:
+    relations = relation_content_types(any_model=type(obj))
     types = set()
-    for relation in relations:
-        types.update(relation.model_class().obj_list())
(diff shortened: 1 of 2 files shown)


apis_core/relations/templatetags/relations.py
```diff
@@ -15,11 +15,14 @@ def possible_relation_types_from(obj) -> list[ContentType]:
...
-    for relation in relations:
-        types.update(relation.model_class().obj_list())
-        types.update(relation.model_class().subj_list())
+    for model in [relation.model_class() for relation in relations]:
+        if type(obj) in model.obj_list():
+            types.update(model.subj_list())
+        if type(obj) in model.subj_list():
+            types.update(model.obj_list())
     return sorted(
         list(map(ContentType.objects.get_for_model, types)), key=lambda x: x.name
```
sennierer, line 27:
```suggestion
        list(map(ContentType.objects.get_for_model, types)), key=attrgetter("name")
```
following the article to "stop using lambda" posted by lupl earlier today ;)

sennierer (approved):

PR #1116, opened 2024-08-28
Ms/1112 history adapt the history view to show also ng relations

sennierer:


b1rger (changes requested):
see additional commit

b1rger (approved):

PR #1119, opened 2024-08-29
feat(generic): add DateTimeIntervalCharField

b1rger:


sennierer:
Is there a specific reason to use `DateTimeField` instead of `DateField`?
I guess our main use case (99%) has only dates and no time.

apis_core/generic/fields.py
```diff
@@ -0,0 +1,106 @@
...
+        return super().pre_save(model_instance, add)
+
+
+DEFAULT_DATE_REGEX = (r"(?P<day>\d{1,2})\.(?P<month>\d{1,2}).(?P<year>\d{1,4})",)
+
+
+class FuzzyDateRegexField(GenericDateIntervalField):
+    def __init__(
+        self,
+        regex_patterns: List[Tuple[str, str, str]] = [
```
sennierer, line 58: as discusssed remove the list here and add a doctring explaining the sequence of patterns: sort, from, to

sennierer (changes requested):

b1rger:
I've reworked the FuzzyDateRegexField as well as the GenericDateIntervalField, which is the base for the regex and the parser fields. I had to redo the `contribute_to_class` method in the `GenericDateIntervalField` because there were, again, problems with the database migration. I've now taken inspiration from a solution from django-money, but it would be great if especially adding/removing fields would be battle tested

sennierer:
Just got a `django.db.utils.OperationalError: duplicate column name: date_of_birth_fuzzy_date_from` when using the new field.
Thats what I did:
- delete the sqlite db 
- add a field called `date_of_birth_fuzzy` with the `GenericDateIntervalField` 
- run `python manage.py makemigrations` 
- run `python manage.py migrate`

Got the error. Checked the migrations and there is no duplicate field definitions in the migrations.

sennierer (approved):
Working as intended. tested with the sample project, both fields. Removed and added fields without any migration issues.

sennierer:
We decided to publish the field as its own module.

koeaw:
> We decided to publish the field as its own module.

Link to repo for reference: https://github.com/acdh-oeaw/django-interval

Issue #1102, opened 2024-08-26
Identify collections by name


b1rger:
Currently, all the templatetags that refer to collections take the id of the collection as argument. This is cumbersome to work with and sometimes ids can change. It would be better, if it was possible to pass the name of the collection. But the name is not distinct. We could make it distinct. Another option would be to use the name + the parent as distinct feature and identify the collection by its full path (i.e. `mytags.myentitytags.foo`)

sennierer:
The latter would be definitely the nicer, but the parent could have a parent...so if we opt for the latter we would need the full path and additionally add a validation logic that sets the full path unique.

b1rger:
> a validation logic that sets the full path unique

I guess if we define in the model that `parent + label` have to be unique, that leads to a unique full path, right?

PR #1121, opened 2024-08-29
birger/make skoscollections unique

b1rger:
- feat(collections): add constraints to collections
- feat(collections): add custom collection manager

Closes: #1102


apis_core/collections/models.py
```diff
@@ -18,6 +33,22 @@ class SkosCollection(GenericModel, models.Model):
...
         ordering = ["name"]
+        constraints = [
+            models.UniqueConstraint(
+                fields=(
+                    "name",
+                    "parent",
+                ),
+                name="unique_name_parent",
+                nulls_distinct=False,
+                violation_error_message="The name and the parent must be unique",
```
sennierer, line 44:
```suggestion
                violation_error_message=\"The combination of name and base collection (parent) must be unique.\",
```
I think the phrasing of the violation message should be a bit clearer.

apis_core/collections/models.py
```diff
@@ -4,6 +4,21 @@
...
+        Return a collection specified by its full path, from the root colletion
+        to the leaf collection, delimited by `|`. I.e. if there is a collection
+        named `foo` and it has a parent named `bar` and `bar` does not have a
+        parent, then you can use the string "bar|foo" to get the `foo` collection.
+        """
+        names = name.split("|")
+        parent = None
+        while names:
+            parent = self.get(parent=parent, name=names.pop(0))
+        return parent
```
sennierer, line 19:
```suggestion
            collection = self.get(parent=collection, name=names.pop(0))
        return collection
```
Might be a minor thing, but the function is actually not returning the parent, but the leaf. So I suggest to change the variable name.

sennierer (approved):

apis_core/collections/models.py
b1rger: I've now used `parent collection` instead of `base collection (parent)` - the term `base collection` is not in use anywhere

apis_core/collections/models.py
b1rger: I understand the concern and its a valid point, but the thing is that it is a `while` loop and the collection item is always the result **and** the parent of the next iteration. Only after the last iteration its not a parent anymore

Issue #793, opened 2024-04-17
enhance the importers to also allow to enrich an already existing entity with data from external entity


sennierer:
eg in Frischmuth we already have places, but are missing lat/long. It would be nice if we could say fetch attributes x/y from external resource.

PR #1137, opened 2024-09-03
birger/793 update external

b1rger:
- **feat(generic): introduce UpdateFromExternal view**
- **feat(apis_entities): link updatefromexternal-view at all linked uris**

Closes: #793

The feature described in #793 was basically implemented in 2e12b7665d4ad717958e3d9f3a03d61caddc4fbf, but this PR also adds the needed interface to be usable for endusers. It depends on the merge view implemented in #1142

### DIFF
--- a/apis_core/apis_entities/templates/apis_entities/partials/linked_open_data.html
+++ b/apis_core/apis_entities/templates/apis_entities/partials/linked_open_data.html
@@ -1,6 +1,13 @@
 <h5 class="card-title">Linked Open Data</h5>
 {% if object.uri_set %}
   <ul>
-    {% for uri in object.uri_set.all %}<li>{{ uri.uri | urlize }}</li>{% endfor %}
+    {% for uri in object.uri_set.all %}
+      <li>
+        {{ uri.uri | urlize }}
+        {% if object.get_change_permission in perms %}
+          <a href="{{ object.get_enrich_url }}?uri={{ uri.uri }}"><span class="material-symbols-outlined">update</span></a>
+        {% endif %}
+      </li>
+    {% endfor %}
   </ul>
 {% endif %}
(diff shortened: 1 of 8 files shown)


apis_core/generic/forms/__init__.py
```diff
@@ -93,3 +93,21 @@ def __init__(self, *args, **kwargs):
...
+        instance = kwargs.pop("instance", None)
+        super().__init__(*args, **kwargs)
+        for key, value in data.items():
+            update_key = f"update_{key}"
+            self.fields[update_key] = forms.BooleanField(
+                required=False,
+                label=f"Update {key} from {getattr(instance, key)} to {value}",
+            )
+
+            self.fields[key] = forms.CharField(initial=value, required=False)
```
sennierer, line 110: Should the kind of field be copied from the original entities form rather than just using `CharField`? What about when the original form adds some consistency checks? or is a Date?

sennierer (approved):
We should also add a button somewhere (maybe in the actions) that allows to enhance (and not merge) the current entity with values from a reference resource. And we should think about whether it makes sense to include the URI of an entity that was used to enhance to the URIs, or maybe allow to set that in a form. The URIs mean `sameAs` othe entity, but if we only include some attributes it could also be that we explicitly dont want to say `sameAs`.

apis_core/apis_entities/templates/apis_entities/partials/linked_open_data.html
```diff
@@ -1,6 +1,13 @@
 <h5 class="card-title">Linked Open Data</h5>
 {% if object.uri_set %}
   <ul>
-    {% for uri in object.uri_set.all %}<li>{{ uri.uri | urlize }}</li>{% endfor %}
+    {% for uri in object.uri_set.all %}
+      <li>
+        {{ uri.uri | urlize }}
+        {% if object.get_change_permission in perms %}
+          <a href="{{ object.get_enrich_url }}?uri={{ uri.uri }}"><span class="material-symbols-outlined">update</span></a>
```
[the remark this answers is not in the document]
b1rger: Created #1257 for that

apis_core/generic/forms/__init__.py
b1rger: consistency checks should be done on the model, not the form

Issue #315, opened 2023-10-04
startup should fail if `APIS_BASE_URI` is not set


b1rger:

koeaw:
Old but thematically related: #19

sennierer:
or maybe use the domain of the running server and overwrite that only if the setting is set

PR #1139, opened 2024-09-03
birger/315 base uri

b1rger:
- **fix(apis_entities): use example.org as fallback and drop global**
- **feat(apis_metainfo): print warning when APIS_BASE_URI is not set**

Closes: #315


sennierer (approved):
Waring etc working as ecpected with sample_project

PR #1177, opened 2024-09-12
birger/cleanup-js

b1rger:
- **refactor(apis_entities): move embedded code to separate file**
- **refactor(apis_entities): drop the `unbind_compare_forms` js function**
- **refactor(apis_entities): drop javascript event trigger**
- **refactor(apis_entities): drop unnecessary javascript method**
- **refactor(apis_entities): drop the csrfSafeMethod javascript method**
- **refactor(apis_entities,core): move javascript multiselect initialization**
- **refactor(apis_entities): drop the apis_templatetag.html template**
- **refactor(core): drop the `bootstrap-select` javascript library**


koeaw (changes requested):
I've looked at this – including rebased onto the latest commit on `main` (2f85686947f31ba28e5807b02ee2fc81cedc62cc) – and it works for me.

I couldn't find references to the removed bits either and the way the`<link>` and `<script>` tags are included in `base.html` is equivalent to what the now removed `apis_templatetag.html` did.

The one thing that needs adding is the missing newline (indicated in comments).

I'd also change the order of the included files so ours are last – to allow overriding of styles and code in dependencies – though this could be done in a follow-up PR.

b1rger:
thanks!

> The one thing that needs adding is the missing newline (indicated in comments).

fixed

> I'd also change the order of the included files so ours are last – to allow overriding of styles and code in dependencies – though this could be done in a follow-up PR.

Good point! I've changed the order

koeaw (approved):

PR #1181, opened 2024-09-12
fix(entities): use field verbose_names in help_text

koeaw:
While trying to update the list of `fields` added to the search field's `help_text` in `ModelSearchFilter` to use `verbose_names` names instead of the fields' names, I ran into an error with how the list of fields is put together.

This PR fixes both how the list is constructed and changes the (default) field names to `verbose_name`s.

### DIFF
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -106,16 +106,19 @@ class ModelSearchFilter(django_filters.CharFilter):
         super().__init__(*args, **kwargs)
 
         if model is not None and "help_text" not in self.extra:
-            if hasattr(model, "_default_search_fields"):
-                fields = model._default_search_fields
+            if default_fields := getattr(model, "_default_search_fields", False):
+                model_fields = [
+                    model._meta.get_field(field) for field in default_fields
+                ]
             else:
                 model_fields = model._meta.fields
-                fields = [
-                    field.name
-                    for field in model_fields
-                    if isinstance(field, (models.CharField, models.TextField))
-                ]
-            fields = ", ".join(fields)
+
+            fields = [
+                field.verbose_name
+                for field in model_fields
+                if isinstance(field, (models.CharField, models.TextField))
+            ]
+            fields = ", ".join(map(str, fields))
             self.extra["help_text"] = f"Search in fields: {fields}"
 
     def filter(self, qs, value):
--- a/apis_core/apis_entities/filtersets.py
+++ b/apis_core/apis_entities/filtersets.py
@@ -109,10 +109,10 @@ class ModelSearchFilter(django_filters.CharFilter):
             if hasattr(model, "_default_search_fields"):
                 fields = model._default_search_fields
             else:
-                modelfields = model._meta.fields
+                model_fields = model._meta.fields
                 fields = [
                     field.name
-                    for field in modelfields
+                    for field in model_fields
                     if isinstance(field, (models.CharField, models.TextField))
                 ]
             fields = ", ".join(fields)


koeaw:
The error I encountered read:
```
TypeError: sequence item 0: expected str instance, __proxy__ found
```

koeaw:
I actually also wanted to convert the field names provided via `_default_search_fields` to `verbose_names`  + update some of the documentation (docstrings, comments), but I'm unclear on how `_default_search_fields` is supposed to work.

A comment which mentions it makes it sound like it's supposed to be a list on the model, but non-field data types aren't really a thing on Django models, or are they? I was thinking it should probably be a class method (?), in which case it would need to be referenced differently.

b1rger (changes requested):
* you're importing `logging` and creating a `logger` without using it. please remove those, adding logging to the filters is a separate change, different from the one you want to implement.
* you're mixing stylistic changes with a functionality change. please don't do that. quote from `CONTRIBUTING.md`: "Please try to keep your commit changes focused on the change you want to implement - don't start fixing typos if your commit is actually about adding a feature."
* `_default_search_fields` is a list of fieldnames that is used for searching. In the current state, the filters help text lists either fieldnames (if `_default_search_fields` is set) or verbose names (if `_default_search_fields` is not set). This should be harmonized to show verbose names in both cases
* The commit message should explain why the switch from `fields` to `get_fields()` is necessary
* Are you sure that `map` to convert the values to `string` is necessary? `verbose_name` is a string, so `fields` is already a list of strings, isn't it?

koeaw:
Fair enough on logging & variable name change – was already preparing to split commits for a follow-up PR for docs off in order to separate concerns.

Regarding the other points:

Re: consistency in returned values: as said, I meant to update the fields from `_default_search_fields` as well, but while trying to do so, I ran into the issue of not understanding how it is meant to be set on the model in the first place. The comment I referenced says:
```python
# check if the model has a `_default_search_fields`
# list and use that as searchfields
```
What does that mean? The Django docs don't make mention of using regular data types as quasi constants. Even if they are technically allowed, attributes set on Django models are expected to point to fields. Extra information goes either into `Meta` or in methods. If `_default_search_fields` was meant to be a method, it isn't currently being referenced correctly. So, confusing either way.

And I disagree that the switch from `fields` to `get_fields()` needs to be explained. The latter is the [standard way](https://docs.djangoproject.com/en/5.1/ref/models/meta/#module-django.db.models.options) of retrieving information about fields...

Also, the string conversion is necessary. Our `verbose_names` are of type `<class 'django.utils.functional.lazy.<locals>.__proxy__'>`, not `str`. They are [translated strings](https://docs.djangoproject.com/en/5.1/ref/unicode/#translated-strings).

b1rger:
> What does that mean? The Django docs don't make mention of using regular data types as quasi constants.

A Django model is a Python class. A Python class can have attributes.

> Even if they are technically allowed, attributes set on Django models are expected to point to fields.

Is there some documentation that states this? We discussed the approach of the `_default_search_fields` at length in the the JFX, I'm a bit astonished that this was not mentioned back then? If thats not allowed, we have to rework this whole approach

> Extra information goes either into `Meta` or in methods. If `_default_search_fields` was meant to be a method, it isn't currently being referenced correctly. So, confusing either way.

We can not put custom attributes into Meta.

> 
> And I disagree that the switch from `fields` to `get_fields()` needs to be explained. The latter is the [standard way](https://docs.djangoproject.com/en/5.1/ref/models/meta/#retrieving-all-field-instances-of-a-model) of retrieving information about fields...

Well, but does it make a difference regarding the change you want to implement? Or is it just a separate unrelated change thats just piggybacked in this commit?

> Also, the string conversion is necessary. Our `verbose_names` are of type `<class 'django.utils.functional.lazy.<locals>.__proxy__'>`, not `str`. They are [translated strings](https://docs.djangoproject.com/en/5.1/ref/unicode/#translated-strings).

Then this is exactly what should be explained in the commit message

b1rger:
> > Even if they are technically allowed, attributes set on Django models are expected to point to fields.
> 
> Is there some documentation that states this? We discussed the approach of the `_default_search_fields` at length in the the JFX, I'm a bit astonished that this was not mentioned back then? If thats not allowed, we have to rework this whole approach

Actually it would be pretty bad if we were violating some rule, given that we also use class attributes that are not fields in the new relations app.

b1rger:
> I don't understand how you arrive at them not being allowed when I said the opposite in the bit you quoted. Not sure if this is a language/communication issue.

I'm not sure what you meant by "Django models are expected to point to fields". This is what I'm referring to. "Technically allowed" does not mean it makes sense to write code that way, especially if it leads to misunderstandings. We can do better. If this is some kind of best practice I would like to know about it so we can adapt our code so its more usable.

koeaw:
> I'm not sure what you meant by "Django models are expected to point to fields". This is what I'm referring to. "Technically allowed" does not mean it makes sense to write code that way, especially if it leads to misunderstandings. We can do better. If this is some kind of best practice I would like to know about it so we can adapt our code so its more usable.

Ah, ok.

I was basing my wording on:
* the Django docs starting out with, "Each attribute of the model represents a database field." [when they first introduce models](https://docs.djangoproject.com/en/5.1/topics/db/models/) (top of the page),
* the docs not addressing or demonstrating meta information about models or "random" data being stored that way,
* the fact "attributes" are rarely mentioned at all – and when they are, it's primarily in the context of:
	* data that feeds *into* fields (like constants for `Choices`) or
	* more "under the hood" stuff (like `objects`/`Managers`).

Online examples of using attributes for other things are also few and far between.

I didn't really give much thought to `_default_search_fields`' implementation details when `ModelSearchFilter` was discussed at the JF because AFAIR we already had our custom search fields at that point and I was more concerned with integrating the new default search fields with the existing ones in `filtersets.py`.

Now I'm reminded of our old approach of storing settings in `APIS_ENTITIES` and how even alternative approaches discussed at the time, like storing fields alongside the models, are cumbersome because 1) of how easily they (accidentally) go out of sync and 2) they require updates to code files and redeployments for the smallest adjustments to business rules.

It seems like `ModelAdmin` is the one built-in thing that deals with models on a "meta" level, but it serves a specific, different purpose – driving the admin UI – and is probably not meant to (or too complex to) have only some of its functionality reused for similar needs.

koeaw:
> * you're importing `logging` and creating a `logger` without using it. please remove those, adding logging to the filters is a separate change, different from the one you want to implement.

Removed.

> * you're mixing stylistic changes with a functionality change. please don't do that. quote from `CONTRIBUTING.md`: "Please try to keep your commit changes focused on the change you want to implement - don't start fixing typos if your commit is actually about adding a feature."

Moved to separate commit.

> * `_default_search_fields` is a list of fieldnames that is used for searching. In the current state, the filters help text lists either fieldnames (if `_default_search_fields` is set) or verbose names (if `_default_search_fields` is not set). This should be harmonized to show verbose names in both cases

Extended fix to use `verbose_names` in both cases.

> * The commit message should explain why the switch from `fields` to `get_fields()` is necessary

> Well, but does it make a difference regarding the change you want to implement? Or is it just a separate unrelated change thats just piggybacked in this commit?

Summarised what the switch is about.

This is a case of repeated (mis)use of Django internals, which severely limits the usefulness/further processability of the resulting code. I don't think it needs continued spelling out in detail, much as I wouldn't expect e.g. every fix for built-in functionality being called with an incorrect argument to always discuss what exactly is wrong with it/how it is incorrect or to mention it by (its old, incorrect) name.

> > Also, the string conversion is necessary. Our `verbose_names` are of type `<class 'django.utils.functional.lazy.<locals>.__proxy__'>`, not `str`. They are [translated strings](https://docs.djangoproject.com/en/5.1/ref/unicode/#translated-strings).
> 
> Then this is exactly what should be explained in the commit message

Gave translation strings (type) as reason for change.

b1rger:
> > * The commit message should explain why the switch from `fields` to `get_fields()` is necessary
> 
> > Well, but does it make a difference regarding the change you want to implement? Or is it just a separate unrelated change thats just piggybacked in this commit?
> 
> Summarised what the switch is about.

Oke, you write in the commit message: "Requires fixing of retrieval of field meta information from Django-internal `fields` method to `get_fields`". But at the same time, in the code you now pushed dropped the switch to `get_fields`. So is it required in the sense, that it does *NOT* work without the switch? At least in my tests I did right now the code worked without the switch. Which makes me think, that this still is a purely stylistic change, that is *NOT* required to make the change work?

> This is a case of repeated (mis)use of Django internals, which severely limits the usefulness/further processability of the resulting code. 

Thanks for the valuable feedback! Its nice to hear that we actually have no idea what we are doing! It would be so great if someone with wisdom would help us write more useful code, by, for example, doing more contributions themselves or doing PR reviews!

> I don't think it needs continued spelling out in detail, much as I wouldn't expect e.g. every fix for built-in functionality being called with an incorrect argument to always discuss what exactly is wrong with it/how it is incorrect or to mention it by (its old, incorrect) name.

If you want someone to review your code, you should describe your changes in the commit message, so the person reviewing the code knows *why* the change happened.

Please change `entities` scope to `apis_entites` - there is no `entitites` app in `apis_core`.

(It would have been great to have an issue where we could discuss this approach beforehand - a similar block of code is also used in the `generate_search_filter` function. If there is a generic method that returns the search fields for a model, that could be used in two places)

b1rger (changes requested):
see comment

PR #1194, opened 2024-09-18
prevents choices being treated as autocomplete

### LINKED ISSUE #1193: CreateRelationsForm - django fields with choices are displayed as autocomplete (without any options)
When a relationship class has a choices field then the following code in ```apis-core-rdf/apis_core/relations/views.py```
converts it into an AutoComplete field but the choices are lost.
```
    def get(self, *args, **kwargs):
        resp = super().get(*args, **kwargs)
        content_type = ContentType.objects.get_for_model(self.model)
        resp["HX-Trigger-After-Settle"] = (
            '{"reinit_select2": "relation_' + content_type.model + '_form"}'
        )
        return resp
```
~I couldn't see any loss in the autocomplete functionality when I removed this method; and~ once I remove this function my choices field also works fine as a dropdown.
 
@b1rger is my understanding correct?

gythaogg:
fixes #1193

### DIFF
--- a/apis_core/relations/templates/apis_core/apis_entities/abstractentity_form.html
+++ b/apis_core/relations/templates/apis_core/apis_entities/abstractentity_form.html
@@ -13,7 +13,7 @@
   });
   document.body.addEventListener("reinit_select2", function(evt) {
     form = document.getElementById(evt.detail.value);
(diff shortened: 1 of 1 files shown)


b1rger:
thanks!

PR #1195, opened 2024-09-18
fix(merge-entities): use source obj's bool value

### LINKED ISSUE #818: change the default behaviour of the `merge_with` function for boolean fields
currently booleanfields are connected via and when merged: 
https://github.com/acdh-oeaw/apis-core-rdf/blob/fcadf02c1301a13051eb84b3114d4a441e705cd0/apis_core/apis_entities/models.py#L121-L124

For boolean fields we want to keep the value in the source object

gythaogg:
fixes #818

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -118,9 +118,7 @@ class AbstractEntity(RootObject):
         setattr(self, field.name, res)
 
     def merge_booleanfield(self, other, field):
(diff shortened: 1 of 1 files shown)


b1rger (changes requested):
Thanks! Could you change the scope from `merge-entities` to `apis_entities`? Currently we use the app name as the commit message scope in most cases

PR #1214, opened 2024-09-25
refactor(apis_metainfo)!: drop unused attributes from Uri model

b1rger:


sennierer (approved):
as discussed these fields have never been used and can be removed

PR #1234, opened 2024-09-27
Redistribute scroll button code

koeaw:
Move CSS and JS for scroll-to-top button from template to `core.css`, `core.js`,
add remaining button HTML back into `base.html` and delete the partial.

Code and markup remain unchanged (for now), only added comments + enclosed JS in event listener.


koeaw:
Needs to be rebased on `main` once #1177 is merged.

b1rger (commented):
Hm, I think it would make sense to have the scroll to top code in separate `js` and `css` files, something like `scrolltotop.js` + `scrolltotop.css`

koeaw:
> Hm, I think it would make sense to have the scroll to top code in separate `js` and `css` files, something like `scrolltotop.js` + `scrolltotop.css`

I think it'd be a good idea to discuss how we generally wanna handle CSS and JS going forward, including how things should get split up/out/off, new blocks in templates that CSS and JS should get wrapped in for easier overriding etc., at a JF.

It'd be easier/faster for me to refactor this again then, to whatever we decide, based on this PR than to redo the untangling. (Also because I already have more changes lined up based on this now. :grimacing:) So if this could be merged now, that'd be great.

koeaw:
Ticket for proposed JF discussion: #1240

sennierer:
js files separated depending on the templates the functions get loaded in. Eg if a function needs to be loaded in several locations it might need to be in a separate file, if several functions get loaded in the base template they can be put in one file.

sennierer:
naming convention:
- module_name + js/css

Issue #1235, opened 2024-09-27
Leftover "column_selector" template? 


koeaw:
I didn't see the `apis_metainfo` template [column_selector.html](https://github.com/acdh-oeaw/apis-core-rdf/blob/main/apis_core/apis_metainfo/templates/apis_metainfo/tags/column_selector.html) referenced anywhere, can it be deleted? Is it still being used somewhere?

It contains JS code which would need splitting out in a separate file – _if_ it's meant to be kept.

koeaw:
Hmm, this looks like it is (was) meant for the column selector (?!) for the table view, which I assumed was based on `boostrap-multiselect `because it looks identical to the dropdowns for fields with choices (which use that).

PR #1236, opened 2024-09-30
chore(apis_metainfo): drop unused template files

b1rger:
Those files were used by templatetags which were removed in
ed1c38dff77d2c0951336a00f5dae1c41cbd8bb6

Closes: #1235


koeaw (approved):
I didn't mention [class_definition.html](https://github.com/acdh-oeaw/apis-core-rdf/pull/1236/files#diff-e468807f62cef878d95ebbaa36ac2bf085225176d08600651704fdbb5bd00c98) in the linked issue because I wasn't sure if we'd intended to add it back in... though if we wanted to reintroduce it, it should probably be made optional.

So, yeah, best to remove it, too, for now.

PR #1237, opened 2024-09-30
fix(apis_relations): edit button appearance

koeaw:
Style edit button in relations tables consistent
with edit button in list view (colour, title).


koeaw:
Thx!

PR #1242, opened 2024-09-30
refactor(apis_entities): object navigation

koeaw:
Replace near-duplicate templates used for list object navigation in detail view (next, previous object) with a single template which is called with/gets rendered based on presence of direction-specific data.


b1rger (commented):
I think this would be an ideal candidate for an [inclusion tag](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#django.template.Library.inclusion_tag).

`icon` is a bit misleading, its the `icon_name` thats usually between the material classed elements.

koeaw:
Changed the variable for the icon to `icon_name`.

Rebased this onto #1262 and refactored positioning from inline styles to CSS classes. Needs rebasing once the other PR is merged -> setting to draft until then.

koeaw:
Closing this for now to maybe revisit at a later point.

Issue #1228, opened 2024-09-26
Move `TempTriple` related code from `apis_entities.models` to a signal in `apis_relations`


b1rger:
we need this to be able to disable `apis_core.apis_relations` in `INSTALLED_APPS`

PR #1254, opened 2024-10-01
fix(apis_entities,apis_relations): move TempTriple update code to signal

b1rger:
Rather than hardcoding the TempTriple update in the merge method, we use
the `post_merge_with` signal to update the existing TempTriples

Closes: #1228


sennierer (approved):
tested this with the ÖBL instance and it works as expected!

PR #1261, opened 2024-10-03
feat(relations): allow more generic relation tables

b1rger:


apis_core/relations/templatetags/relations.py
```diff
@@ -58,11 +58,16 @@ def relations_from(from_obj, relation_type: ContentType = None):
...
+    suffixes = ["RelationsTable"]
+    if target:
+        suffixes.extend(
+            f"{module[-1]}RelationsTable" for module in mro_paths(target.model_class())
+        )
+    table_modules = ()
+    for suffix in suffixes:
+        table_modules += module_paths(
+            type(context["object"]), path="tables", suffix=suffix
+        )
```
sennierer, line 70: could we put that \"find the table class\" functionality in a dedicated function and add a cache?

sennierer (approved):
works as intended

Issue #1265, opened 2024-10-07
Cannot create new relations


gythaogg:
Popups are blocked by the following JS error 
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "default-src 'self' 'unsafe-inline' cdnjs.cloudflare.com cdn.jsdelivr.net fonts.googleapis.com ajax.googleapis.com cdn.rawgit.com *.acdh.oeaw.ac.at unpkg.com fonts.gstatic.com cdn.datatables.net code.highcharts.com *.acdh-dev.oeaw.ac.at *.acdh.oeaw.ac.at openstreetmap.org *.openstreetmap.org".

    at new Function (<anonymous>)
    at htmx.org@1.9.10:1:24672
    at Tr (htmx.org@1.9.10:1:35877)
    at HTMLAnchorElement.a (htmx.org@1.9.10:1:24647)
(anonymous) @ htmx.org@1.9.10:1
Tr @ htmx.org@1.9.10:1
a @ htmx.org@1.9.10:1Understand this error

```
Related to https://github.com/acdh-oeaw/apis-instance-nomansland-dev/issues/14

Update from @b1rger 

> Introduced in https://github.com/acdh-oeaw/apis-core-rdf/commit/a2e80fad3fcfc2903067a072de13af4d0406b
> Oke, long debugging session later: as the error messages says, the problem is CSP, which enabled in our default settings, but which is not enabled in the discworld sample project. Apparently I did only test this commit with the sample project. For now it helps to add 'unsafe-eval' to the CSP_DEFAULT_SRC, but we should try to find a better fix. Therefore I suggest to only do this in the projects and not in the default settings.

PR #1270, opened 2024-10-07
fix(relations): use javascript onclick instead of htmx on:click

b1rger:
This makes the trigger also work with a strict CSP that does not allow
`unsafe-eval`.

Closes: #1265


b1rger:
Could you try with Django 5?

PR #1272, opened 2024-10-08
feat(generic): more/less text column template

### LINKED ISSUE #944: preview with more/less link in list view for columns with lots of text
Currently used in Tibschol - create a PR for apis-core-rdf

gythaogg:
This pull request introduces a new `MoreLessColumn` feature to handle long text fields in tables, providing a preview with a "Show more" link that expands to reveal the full text. The changes span across CSS, template, and Python files to implement this functionality.

### New `MoreLessColumn`:

* **CSS for MoreLessColumn**:
  * Added styles in `apis_core/generic/static/css/more-less-column.css` to handle the appearance and behavior of the "Show more" and "Show less" links.

* **Template for MoreLessColumn**:
  * Created a new template `apis_core/generic/templates/columns/more-less.html` to render the preview and full text within a `<details>` element.

* **Python class for MoreLessColumn**:
  * Added `MoreLessColumn` class in `apis_core/generic/tables.py` to use the new template and CSS for rendering long text fields with expandable previews.

* **Usage example**:
  * Updated `sample_project/tables.py` to include an example of how to use the `MoreLessColumn` in a `PersonTable`.
 
Closes #944

### DIFF
--- a/sample_project/tables.py
+++ b/sample_project/tables.py
@@ -1,0 +1,14 @@
+from apis_core.apis_entities.tables import AbstractEntityTable
+from apis_core.generic.tables import MoreLessColumn
+
+
+class PersonTable(AbstractEntityTable):
+    class Meta(AbstractEntityTable.Meta):
+        sequence = ["desc", "bio"]
+
+    # example column to show preview and detail
+    bio = MoreLessColumn(
+        orderable=False,
+        preview=lambda x: f"This is {x.surname}",
+        fulltext=lambda x: f"This is {x.surname}, {x.forename} {x.surname}.",
+    )
--- a/apis_core/generic/static/css/more-less-column.css
+++ b/apis_core/generic/static/css/more-less-column.css
@@ -1,0 +1,42 @@
+summary.more-less-column {
+  list-style: none;
+  appearance: none;
+  -webkit-appearance: none; /* for WebKit browsers */
+}
+
+/* optional: remove ::marker for older browsers */
+summary.more-less-column ::marker {
+  content: none;
+}
+
+summary.more-less-column {
+  cursor: pointer;
+}
+
+/* add "Show more" text when collapsed */
+summary.more-less-column::after {
+  content: " Show more";
+  font-weight: normal;
+  font-size: smaller;
+  color: grey;
+}
+
+/* hide the summary text when expanded, show only "Show less" */
+details[open] summary.more-less-column::before {
+  content: "Show less";
+  color: grey;
+}
+
+/* hide the original summary text when expanded */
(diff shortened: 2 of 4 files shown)


koeaw:
Not a reviewer, but if we do use separate static files for functionality, I think their names shouldn't be too generic. E.g. "preview" alone is rather nondescript, sth. like "preview-column" or "table-preview" or similar (off of the descriptions; haven't looked at the thing yet) would be more specific.

sample_project/tables.py
b1rger, line 1: I think this is a great column and I know at least one project that will make use of it!

I think we can make the use of the column a bit more "typical" to how django tables columns are used. Currently I have to use a `render_` method on the field, but if I do that, I can also just use a `TemplateField` directly, pass the template name and pass the additional context parameters in the `render_` method.
But we can make it easier for the users and provide a field that does **not** require to define a custome `render_` method - by allowing to pass arguments to `MoreLessColumn` that return `preview` and `fulltext`:
```

class MoreLessColumn(tables.TemplateColumn):
    """
    Useful for displaying long fields,
    a preview is shown initially with a Show more link
    which is replaced with a Show less link when expanded
    """

    template_name = "columns/more-less.html"
    empty_values = ()

    def __init__(self, preview, fulltext, *args, **kwargs):
        self.preview = preview
        self.fulltext = fulltext
        super().__init__(template_name=self.template_name, *args, **kwargs)

    def render(self, record, **kwargs):
        self.extra_context["preview"] = self.preview(record)
        self.extra_context["fulltext"] = self.fulltext(record)
        return super().render(record, **kwargs)
```

This would make it possible to use the Column like this:
```
bio = MoreLessColumn(orderable=False, preview=lambda x: f"This is {x.surname}", fulltext=lambda x: f"This is {x.forename}")
```
which is similar to how the `linkify` argument of django tables works.

b1rger (changes requested):

b1rger (approved):

b1rger:
I adapted the commit message a bit to make it more verbose
I leave it open for now to give others the chance to test it, but will merge before the next release

koeaw:
@gythaogg I'm removing myself as reviewer because I've run out of hours to test this – already anticipated this, reason why I didn't volunteer, sry!

The only other _stylistic_ feedback I have is that I'd a) change comment blocks to start with lower case letters for consistency & because new words can seamlessly be added to the front. And b) Docstrings needn't be as curt and can be more structured, so would break that into two proper sentences with full stops at the end and put the UI labelling/wording mentioned ("Show more"...) in quotation marks for easier parsing.

PR #1276, opened 2024-10-08
feat(relations): list relations under tabs

### LINKED ISSUE #489: Redesigning relations list view for entity

gythaogg:
This pull request includes an update to the `list_relations_include.html` template to enhance the user interface by switching from a card layout to a tabbed navigation layout.

UI Improvements:

* [`apis_core/relations/templates/relations/list_relations_include.html`](diffhunk://#diff-d04981e387d63223046d9453920a5aa0805ed95e4532e75e8261c074bb132adfR5-R20): Replaced the card layout with a tabbed navigation layout to improve the user experience. The new layout uses Bootstrap's tab component to organize relation targets into tabs.

closes #489 

TO DISCUSS

- [x] Potentially useful to have an ALL tab to have a quick overview and set that as default. ~This is not implemented in this pull request.~
- [ ] TibSchol users would prefer the tab headings to also show the number of relations under each tab so they don't have to cycle through all the tabs to see which one has content. Don't know if it is worth it (it was only a nice-to-have wish). 
- [ ] When you are in a particular tab and then do an action that causes the page to reload (by clicking on Edit, for example) then the first tab will be active again. 

~When new relations are created, they should be added to the correct element so that they are not displayed under all tabs~

After new relations are created, they appear correctly in the active tab but need a refresh for them to appear in the ALL tab.

### DIFF
--- a/sample_project/templates/sample_project/person_detail.html
+++ b/sample_project/templates/sample_project/person_detail.html
@@ -1,0 +1,5 @@
+{% extends "apis_core/apis_entities/abstractentity_detail.html" %}
+
+{% block relations-include %}
+  {% include "relations/list_relations_include_tabs.html" %}
+{% endblock relations-include %}
--- a/sample_project/templates/sample_project/person_form.html
+++ b/sample_project/templates/sample_project/person_form.html
@@ -1,0 +1,5 @@
+{% extends "apis_core/apis_entities/abstractentity_form.html" %}
+
+{% block relations-include %}
+  {% include "relations/list_relations_include_tabs.html" with edit=True %}
+{% endblock relations-include %}


b1rger (commented):
Nice! We should discuss in JFX if this should be the default. An "ALL" tab would definitly make sene.

Regarding the problem with all relations being listed after a change: do you already know what we need to change to fix that? Mabye we have to refactor the `list_relations.html` template a bit and pass the target as an argument...

sennierer:
look into keeping both options
also see if anchor elements plus scrolling and rolling headers makes sense
also sorting of tables needs to be sorted

b1rger:
> @b1rger I tried to override the person template to use the tabbed view of relations for sample_project - [3f75d7f](https://github.com/acdh-oeaw/apis-core-rdf/commit/3f75d7f269785fe6f43b5960678e74fb98e8c361)
> 
> It looks like it displays the relations in tabs and via the block.super call, it also displays the relations in the grouped list format and I can't quite figure out how to skip the overridden template in `relations/templates/apis_core/apis_entities/... `and only run the super code in `apis_core/apis/entities/`
> 
> View image
> 
> Edit image
> 
> Am I doing something wrong here?

Sorry, I'm not quite sure what you want to achieve. Why don't you drop the `{{ block.super }}`?

apis_core/relations/templates/relations/list_relations.html
b1rger, line 1: What is this change for?

b1rger (changes requested):
* Please move the introduction of the `relations-include` block into a separate PR
* I'm not sure what the "To Be Done" in one of the commit messges means - does it mean that this PR is still a draft and those things will be implemented? Or should those todos be implemented after the PR is merged?

b1rger:
> Without this change if I override col-one in my ontology (to point to the tabs relation template,- in the inheritance chain the call to block.super includes the col-one from the templates in relations first - which includes default listing of relations.

I'm not sure if we talk about the same change. I'm talking about the addition of the `ALL` if no target exists.

apis_core/relations/templates/relations/list_relations_include_tabs.html
```diff
@@ -0,0 +1,28 @@
...
+  {% endfor %}
+</ul>
+<div class="tab-content">
+  <div class="tab-pane container active" id="reltab_{{ object.id }}_ALL">
+    {% include "relations/list_relations.html" %}
+  </div>
+  {% for target in possible_targets %}
+    <div class="tab-pane container"
+         id="reltab_{{ object.id }}_{{ target.name }}"
+         hx-swap-oob="true">{% include "relations/list_relations.html" %}</div>
```
b1rger, line 26: There is still a `hx-swap-oob` leftover

b1rger (changes requested):

b1rger (approved):

PR #1283, opened 2024-10-11
fix(apis_entities): correct merge link

### LINKED ISSUE #1278: bug(apis_entities): merge link broken
Steps to reproduce:
1. Go to any entity detail page (Example: https://discworld.acdh-ch-dev.oeaw.ac.at/sample_project.person/1)
2. Click on **Actions** and then **Merge** in the Actions menu

Expected behaviour: A form to select to the entity with which this entity needs to be merged
Actual behaviour: The page reloads.

I think this was introduced in https://github.com/acdh-oeaw/apis-core-rdf/commit/ba7ad4c2b450e83069b960d012ff4e3f544306e8 when `get_merge_url` was removed from `apis_entities.models` which used return `generic_entities_merge_view`

The `get_merge_url` in `generic/abc.py` (introduced in https://github.com/acdh-oeaw/apis-core-rdf/commit/5f21c4dfb1a428ff076006029f6474c569c9fa1a) instead returns `apis_core:generic:merge` which shouldn't be called on clicking the Merge link

@b1rger: It might make sense to remove Merge from this Action menu and use the Merge field in the Edit view below the Relations table

gythaogg:
The merge link was incorrectly pointing the merge action view that merges two objects; it has been corrected to point to the merge view that allows the user to select another object for merging.

closes #1278

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -84,6 +84,13 @@ class AbstractEntity(RootObject):
             kwargs={"contenttype": entity, "pk": self.id},
         )
 
+    def get_merge_view_url(self):
+        entity = self.__class__.__name__.lower()
+        return reverse(
+            "apis_core:apis_entities:generic_entities_merge_view",
+            kwargs={"contenttype": entity, "pk": self.id},
+        )
+
     def merge_start_date_written(self, other):
         self.start_date_written = self.start_date_written or other.start_date_written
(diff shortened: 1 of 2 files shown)


sennierer (approved):

b1rger:
(I've moved the method to the `AbstractEntity`, given that the url is also defined in the `apis_entity` module)

PR #1301, opened 2024-10-21
refactor(apis_relations): move external methods into module

b1rger:
This commit moves a couple of methos and templatetags into the
`apis_relations` module, that were previously spread over the `utils`
and the `apis_entities` module. Having them contained in the
`apis_relations` module moves us one step further in being able to allow
disabling the `apis_relations` module.


sennierer (approved):
working with discworld and oebl as intended

PR #1338, opened 2024-11-04
relations: make autocomplete use specific endpoint

b1rger:
* fix(relations): if there is only one target, use specific autocomplete
When the relation form only has to provide autocomplete for a specific
target type, we can use the generic autocomplete endpoint. This allows
also to create entities by pasting URIs.

* fix(relations): force relation form select2 to use html

### DIFF
--- a/apis_core/relations/static/js/relation_dialog.js
+++ b/apis_core/relations/static/js/relation_dialog.js
@@ -1,3 +1,8 @@
+function tohtml(item) {
+    const span = document.createElement('span');
+    span.innerHTML = item.text;
+    return span;
+}
 document.body.addEventListener("reinit_select2", function(evt) {
     form = document.getElementById(evt.detail.value);
     form.querySelectorAll(".listselect2").forEach(element => {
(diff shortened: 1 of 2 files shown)


apis_core/relations/static/js/relation_dialog.js
```diff
@@ -1,3 +1,8 @@
+function tohtml(item) {
+    var $result = $('<span>');
+    $result.html(item.text);
+    return $result;
+}
```
sennierer, line 5:
```suggestion
function tohtml(item) {
    const span = document.createElement('span');
    span.innerHTML = item.text;
    return span;
}
```
given that we want to get rid of jquery, use plain JS instead of jquery for the formatting

sennierer (changes requested):
The metadata is now working nicely, but its still not possible to paste urls to the select as it is possible in the import form.
Please change the jquery code to plain JS as suggeste.

sennierer (approved):

PR #1345, opened 2024-11-05
feat(sample_project): annotated queryset for Place

gythaogg:
Place queryset is annotated with two fields called resident_id, and resident_name that picks one resident (person) living in that place using the LivesIn relationship.

reproduces issue described in #1346


b1rger:
should we close this, now that the issue is resolved? or do you want to have this reproducer in the `sample_project`?

Issue #1343, opened 2024-11-05
Accessibility issues with links/icons in footer


koeaw:
The `title` attributes on the icon links in the footer should get removed/replaced with `alt` attributes on the images.

That aside, I'm experiencing weird behaviour when hovering the icons: the link destinations aren't shown consistently. In FF (ESR), I thought they randomly disappeared, then realised they flip between the bottom left corner of the browser (default) and the bottom right. In Chrome, they drop "under" the visible browser window. Additionally, the space between the icons seems to get underlined on hover (like text would by default).

A cursory inspection revealed the links themselves aren't the same size as the icons they surround; I'm assuming this contributes to or is the source of the issue.

b1rger:
> The `link` attributes for the icon destinations in the footer should get removed/replaced with `alt` attributes on the images.

Can you post an example? I don't know what you are referring to

koeaw:
> Can you post an example? I don't know what you are referring to

Sorry, meant to say `title` attributes! Updated the issue description.

b1rger:
Oke, could you explain why it is preferable to have `alt` attributes instead of `title` attributes?

koeaw:
> Oke, could you explain why it is preferable to have `alt` attributes instead of `title` attributes?


The [HTML specs on `title`](https://html.spec.whatwg.org/multipage/dom.html#the-title-attribute):

> Relying on the `title` attribute is currently discouraged as many user agents do not expose the attribute in an accessible manner as required by this specification (e.g., requiring a pointing device such as a mouse to cause a tooltip to appear, which excludes keyboard-only users and touch-only users, such as anyone with a modern phone or tablet).

The [HTML specs on `alt` in the context of links containing nothing but an image](https://html.spec.whatwg.org/multipage/images.html#a-link-or-button-containing-nothing-but-the-image): 

> When an `a` element that creates a hyperlink, or a `button` element, has no textual content but contains one or more images, the `alt` attributes must contain text that together convey the purpose of the link or button.

Useful resources which link to more useful resources:
https://www.a11yproject.com/posts/title-attributes/ 
https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

b1rger:
The one issue was fixed in cfa9006472d59ad8e544173b85cb2bf7d28a2ffa, the other in c2c5594cc73dd61eb5b6a42aa402c2c3d98bec6b

PR #1347, opened 2024-11-05
fix(core): fix footer link dimensions

koeaw:
Set display property of links in site footer
which are visualised by icons – which contain
images or <span> tags for Material Symbols –
to inline-block to make them fit the icons'
dimensions. Prevents links that are smaller
in size than their visual content, which can
cause unexpected behaviour when hovering.

Solves one of the issues described in #1343.

### DIFF
--- a/apis_core/core/static/css/core.css
+++ b/apis_core/core/static/css/core.css
@@ -7,6 +7,11 @@ a #logo:hover {
     opacity: 0.5;
 }
 
+footer a:has(> img),
+footer a:has(> span[class*="material-symbols"]) {
+    display: inline-block;
+}
+
 footer a:has(> img):hover,
 footer a:has(> span[class*="material-symbols"]):hover {
     text-decoration: none;


b1rger:
Can you explain in a bit more detail what this does? "unexpected behaviour when hovering" does not really explain what the problem is.

I tested the PR and I can not see any difference when interacting with the footer elements

koeaw:
> Can you explain in a bit more detail what this does? "unexpected behaviour when hovering" does not really explain what the problem is.

From the issue:

> A cursory inspection revealed the links themselves aren't the same size as the icons they surround; I'm assuming this contributes to or is the source of the issue.

When the size and/or positioning of containing and contained elements don't match, spillovers, overlays, offsets or areas without overlap can be the consequence. Users may not necessarily be aware of those until the elements are interacted with.

If a link does not cover the same area as its content, users may end up hovering/clicking/... in the wrong place. When tabbing, the border on the focussed element may look off. Etc.


footer_icons_link_mismatch

I can reword the commit message to be more neutral, e.g. change it to "Prevents links that are smaller in size than their visual content, which can cause unexpected behaviour **when interacting with them**.", but please let's not turn our frontend commits into frontend basics explainers. This commit message is already 3 times the length of the change itself.

b1rger:
I'll merge this once I can reproduce the issue

koeaw:
> Changes look good to me.

Thx.

> Perhaps in the commit message you meant "focus" instead of "hover" - that's when I would see a box around the link,

I originally noticed on mouse hover – the link destinations would pop up/disappear in areas not 100% consistent with the images. The alternative wording I suggested about would cover any of these without being specific about them.

koeaw:
Changed the last sentence to (slightly modified from above):
"Prevents links that are smaller in size than their visual content, which can cause unexpected behaviour **when they are interacted with.**"

koeaw:
> I'll merge this once I can reproduce the issue

Ok.

Issue #1331, opened 2024-10-29
GIT_REPOSITORY_URL logo


koeaw:
I'd like to suggest to change the logo used for `GIT_REPOSITORY_URL` from the [Git](https://git-scm.com/book/en/v2) logo (favicon for the official Git documentation, Git book) to one that reflects the platform that code is hosted on, in our case: the GitHub logo.

This reflects icon usage for code hosting platforms/linkage to repos across the web.

If we want to use a generic fallback, maybe we can find a logo that's not already associated with one very specific website, i.e. the one linked above, as that's confusing.

koeaw:
The not-orange version of the Git logo would already be somewhat better than the orange one.

koeaw:
Maybe we can find a variation on the theme of nodes/lines, which techies will recognise as representation of code/commits but which can't be confused with a particular tool or website. [Non-free example](https://www.alamy.com/git-branch-icon-line-isolated-on-white-background-black-flat-thin-icon-on-modern-outline-style-linear-symbol-and-editable-stroke-simple-and-pixel-p-image502916449.html).

b1rger:
> to one that reflects the platform that code is hosted on

we can not know on which platform a project that uses `apis-core-rdf` is hosted on

> This reflects icon usage for code hosting platforms/linkage to repos across the web.

feel free to cite your sources

> If we want to use a generic fallback, maybe we can find a logo that's not already associated with one very specific website, i.e. the one linked above, as that's confusing.

the logo is the official git logo. its associated with the website you linked to, because thats the official git website.

koeaw:
> we can not know on which platform a project that uses apis-core-rdf is hosted on

True. Neither can we know if such a project ultimately uses Git for version control.

[...]

PR #1349, opened 2024-11-05
fix(core): remove title attributes, add alt attributes

koeaw:
Remove (as per the HTML spec) discouraged `title` attributes from links in footer template and instead use their values for missing (also per the HTML spec) `alt` attributes on images they enclose.

Solves one of the issues described in https://github.com/acdh-oeaw/apis-core-rdf/issues/1343.

### DIFF
--- a/apis_core/core/templates/partials/footer-left.html
+++ b/apis_core/core/templates/partials/footer-left.html
@@ -9,12 +9,12 @@
 {% url "apis_core:swagger-ui" as swagger_ui %}
 {% if swagger_ui %}
   <a href="{{ swagger_ui }}" title="Swagger UI">
-    <img src="{% static "/img/Swagger-logo.png" %}" height="24px">
+    <img src="{% static "/img/Swagger-logo.png" %}" alt="Swagger UI" height="24px">
   </a>
 {% endif %}
 {% git_repository_url as repository_url %}
 {% if repository_url %}
   <a href="{{ repository_url }}" title="Git repository">
-    <img src="{% static "/img/Git-Icon-1788C.png" %}" height="24px">
+    <img src="{% static "/img/Git-Icon-1788C.png" %}" alt="Git repository" height="24px">
   </a>
 {% endif %}


b1rger:
Oke, so without the `title` attribute there is now no information about what the icons represent, this was very informative. What do you suggest we use instead?

koeaw:
> Oke, so without the `title` attribute there is now no information about what the icons represent, this was very informative. What do you suggest we use instead?

I'm not sure I understand. The idea of icons (symbols/pictograms) is to convey meaning. The textual description exists for those who can't access the visual; I moved those to the images' `alt` attributes, unchanged, which is where they serve that purpose much better.

If the symbols aren't clear enough, which is why I [flagged the Git logo](https://github.com/acdh-oeaw/apis-core-rdf/issues/1331), we shouldn't use them. One solution is to find better (more obvious, less ambiguous) symbols, another is to use link text.

b1rger:
> One solution is to find better (more obvious, less ambiguous) symbols, another is to use link text.

and a third is to show descriptive information when hovering those symbols or text. and the benefit of this approach is, that it can be combined with any of the other two!

koeaw:
> `alt` won't show a tooltip on hover, right? I believe it is meant to be an alternative when the image doesn't load at all or for screen readers to know what the image is. So @b1rger do you suggest we keep the title; and @koeaw is it okay for title to remain alongside alt?

The HTML spec itself says that use of `title` is discouraged. Please see the links I provided [here](https://github.com/acdh-oeaw/apis-core-rdf/issues/1343#issuecomment-2457150510), including coverage of the topic by A11y experts (going back years).

Using both attributes in parallel has its own problems, not sure if those are discussed in the articles I linked or in resources they, in turn, point to.

So no, I don't think that keeping `title`s for tooltips for visuals which ought to be self-explanatory outweighs the negatives.

koeaw:
> `title` attribute here serves a purpose (tooltip) and hence it should remain. Please leave `title` with the same value as `alt` (as suggested by the [a11yproject](https://www.a11yproject.com/posts/title-attributes/)).

Commit updated to only _add_ `alt` attributes to `` elements, reusing text used in `title` attributes.

Commit message updated to reflect change, additionally replaced mentions of "icons" with "images" for more accuracy re: affected elements. CC type in subject changed to `feat`.

PR #1351, opened 2024-11-07
relations form vertical scroll

gythaogg:
This pull request includes updates to the CSS for the relation dialog and the addition of a new form in the `sample_project` application.

CSS updates:

* [`apis_core/relations/static/css/relations.css`](diffhunk://#diff-9a78191ab3874a83833c754e6d0b09bbe0d6568ad186fa71c96a1e594cb33bd8L9-R10): Changed the `overflow` property of the `#relationdialog` to `overflow-x: visible` and `overflow-y: scroll` for better handling of content overflow.

Form additions:

* [`sample_project/forms.py`](diffhunk://#diff-956ce72449b150c878f635fd4e48f316c1ad921d1bbb66d2bd83b9b0f867474bR1-R6): Added a new form class `LivesInForm` that inherits from `RelationForm` and includes a `notes` field with a large textarea widget.

### DIFF
--- a/apis_core/relations/static/css/relations.css
+++ b/apis_core/relations/static/css/relations.css
@@ -6,7 +6,8 @@
 #relationdialog {
     border: 1px solid black;
     border-radius: .3rem;
-    overflow: visible;
+    overflow-x: visible;
+    overflow-y: scroll;
     padding: 0;
     opacity: 1;
     transform: scale(1);


b1rger (changes requested):
Thanks! That bug already annoyed me and we also have a bug report in oebl for that https://github.com/acdh-oeaw/apis-instance-oebl-pfp/issues/354 !

I'm not sure if it is worth it to add the additional fields to the sample project, but that is definitely up for discussion. If we try to showcase every css fix in the sample project it will soon get overly complex and IIRC the idea of the sample_project was also for external users to have an easily understandable test project they can use to get to know APIS.
Especially in this case, the suggested change did not lead to the wanted showcasing of the bug, because my screen is big enough so that the added form fields also visible without the css fix.

An alternatative to the change of the models (which also bugs me a bit, because it changes the datamodel and adds migrations) would be to override the default form for a relation:
```
from django import forms
from apis_core.relations.forms import RelationForm


class LivesInForm(RelationForm):
    notes = forms.CharField(widget = forms.Textarea(attrs={"rows": "70"}), required=False)
```

Also: could you make the commit message a bit more verbose (I suppose the PR message was generated, but some of that information could be part of the commit message, i.e. "handle overflow separately for the x and y axes, making the y-axis scrollable").
"vertical scroll in relations dlg" is a bit scarce, maybe "handle overflow in relation dialog" and then in the message describe the problem and "handle overflow separately for the x and y axes, making the y-axis scrollable" from the PR description.

In general I think the commit message should be the primary source for information about the changes, not the PR description.

b1rger (approved):
(added a comment to the form and wrote out "dialog" in the commit message)

PR #1360, opened 2024-11-12
fix(relations): fix modal overflow to access dropdown

### LINKED ISSUE #1359: Autocomplete select in new relations modal dialog
Fix #1351 changed the overflow behaviour to scroll; as a consequence when the modal dialog is smaller than the autocomplete select dropdown - the select dropdown gets hidden inside the dialoag with only one option visible like shown below. 
Scrolling the new relations form is also tricky in this case - often leading to accidentally closing the modal window.
image

Ideally it should be like
image

gythaogg:
reverted modal dialog overflow to `visible` in the new relations form so that when the dialog height is shorter than the dropdown, the dropdown remains accessible without having to scroll the dialog.

introduced max-height and auto scroll option for `#relation-dialog-content` so that when the contents of the dialog overflows the dialog height, the dialog is scrollable (maintains the behaviour of fix #1351)

closes #1359

### DIFF
--- a/apis_core/relations/static/css/relations.css
+++ b/apis_core/relations/static/css/relations.css
@@ -6,8 +6,7 @@
 #relationdialog {
     border: 1px solid black;
     border-radius: .3rem;
-    overflow-x: visible;
-    overflow-y: scroll;
+    overflow: visible;
     padding: 0;
     opacity: 1;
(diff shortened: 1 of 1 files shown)


b1rger (approved):
Good catch!

PR #1382, opened 2024-11-13
feat(relations): graph visualisation of relations

gythaogg:
Adds a Graph tab to the tabbed relations template. Graph tab shows a graphical representation of the all relations that include the current object.
Nodes represent objects, each entity type deterministically gets a unique colour.
All edges are directed from the object to the related objects.


sennierer:
@gythaogg whats the status of that PR?

PR #1398, opened 2024-11-15
fix(utils): raise exception on failing creation of object

b1rger:


b1rger:
> But in this case when a user enters a non URI string or if the field is not setup to create objects from external sources - as I guess is the case with Profession in Discworld, could we maybe also inform that it is not possible to create new values from any string and it is currently only possible from URIs, if that is supported? Or in the case that there is no importer defined, could we disable the Create option?

Hm, currently the `GenericImporter` is a fallback for the case when no `Importer` is defined for a model, which means there is always in `Importer` defined. I'd say the Exception message is generic enough to be true for any case (be it a URL that could not be parsed or a non-url string) and we should probably look into ways to enable creating model instances from strings.

PR #1404, opened 2024-11-18
refactor(relations): put display logic of relations in a block

gythaogg:
this way `list_relations_include_tabs` can inherit from the `list_relations_include` template and not contain redundant code to fetch the relations data

@b1rger the large diff is not my fault, it's djlint, my change is only 3 lines :-0 

image


b1rger (changes requested):
please use `feat` - this **adds** the feature of the block (i sometimes also split the `adding` and the `usage` of the block into separate commits)

b1rger (approved):

Issue #1307, opened 2024-10-22
add format prefix from renderer to queryset discovery


sennierer:
similar to the option for serializers implemented in https://github.com/acdh-oeaw/apis-core-rdf/commit/0bcd61bc29cc9c571a3b01319477b5a80bb54f48 but for querysets

sennierer:
not needed anymore, closing for now.

PR #1435, opened 2024-11-22
feat(generic): allow apiview to choose queryset based on renderer

b1rger:
Add additional options to the list of possible querysets, based on the
renderer chosen. This allows users to define separate querysets for
separate renderers.

Closes: #1307


b1rger:
Closing, because issue was closed

Issue #512, opened 2023-12-19
Make `URI` more generic


b1rger:
Uri has a ForeignKey to `RootObject` - it would probably make sense to replace this with a generic foreign key to whatever

sennierer:
Move to generic fk and implement a method on rootobject that allows to keep the way URIs are accessed now (using FKs)

PR #1463, opened 2024-11-29
birger/512 uri

b1rger:
- **feat(apis_metainfo): add a GenericForeignKey to the Uri model**
- **feat(apis_entities): update AbstractEntity to work with generic uri**
- **feat(apis_relations): update GenericTripleForm to work with generic uri**
- **feat(generic): update Enrich view to work with generic uri**
- **feat(utils): update create_object_from_uri to work with generic uri**

Closes: #512


b1rger:
> Data migration works ✔️

thats the most important part ;)

> 
> It seems like this PR removes the feature that allows user to link a Uri to a RootObject 
> Is that tabled for another PR or are we going to do away with manual linking of entities with URIs altogether and only rely on Merge and Import views to add an external Uri to an Object?

good point. I had added `editable=False` to the fields, but to keep the previous behavior it makes more sense to keep them editable. the edit and create forms now allow to set the content_type and object id.

> And likewise, it is possible to link Uri to any object, but at the moment this is also not supported via the webapp, right?

thats now possible

b1rger:
Added a signal to delete URIs of deleted objects and created https://github.com/acdh-oeaw/apis-core-rdf/issues/1516 for the discussed history functionality of URIs

PR #1474, opened 2024-12-04
birger/update multiselect

b1rger:
- **feat(core): update bootstrap-multiselect dependency**
- **feat(core): add additional settings of bootstrap-multiselect**

### DIFF
--- a/apis_core/core/static/js/core.js
+++ b/apis_core/core/static/js/core.js
@@ -2,7 +2,11 @@
 $(document).ready(function() {
     $('select.selectmultiple').multiselect({
         includeSelectAllOption: true,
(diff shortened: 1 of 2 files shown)


b1rger:
> Can confirm that case insensitive filtering works where it didn't work before (in the column select of list views, for example).
> 
> Unable to see any difference in alignment - if you can guide me on how to test it then this can be merged! Thanks!

There should be a better alignment in the column selector.

This is only about the bootstrap-multiselect library. The screenshots you posted are showing the autocomplete select2 widget, which is not affected by that.

Before:
Screenshot-2024-12-05-093955

After:
Screenshot-2024-12-05-094029

PR #1496, opened 2024-12-12
fix: change SPARQL to account for negative lat/long

### LINKED ISSUE #1495: E53_PlaceFromDNB.toml definition doesnt catch all latitude and longitudes
when long/lat is negative the regex doesnt catch it correctly

sennierer:
resolves #1495


b1rger:
Please add scope `(apis_entities)` to the commit message

PR #1504, opened 2024-12-18
feat (documentation): declutter and enrich the datamodel graph

gythaogg:
1. show only one edge between a subject model and an object model
2. show reverse name of the relations in the edge label if it is different from the name
3. slightly reduce the size of the label on the edge

image

TBD:

- [ ]  1 & 2 above  only implemented for new relations, can we drop support for apis_relations?
- [ ]  Since we are showing both forrward and reverse relations, should the edges be bidirectional?

### DIFF
--- a/apis_core/documentation/utils.py
+++ b/apis_core/documentation/utils.py
@@ -36,12 +36,29 @@ class Datamodel:
 
         for rel in self.relations:
             for subj_class in rel.model_class().subj_list():
+                subj_name = ContentType.objects.get_for_model(subj_class).name
                 for obj_class in rel.model_class().obj_list():
-                    key = (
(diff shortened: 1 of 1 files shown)


b1rger (changes requested):
Sorry for the delay.

Regarding the code: i think it makes sense to use `force_str` for all the `name()` methods, maybe do this once, assign the result to variables and then use them in the if/else statement.

Regarding functionality: I think its a feature to be able to distinguish between relations that go in one direction and relations that go in the other direction. The change does show me the reverse name, which is great, but I loose the information in which direction the relation goes (which might be important, because its not always possible to deduce this from the name/reverse_name)

b1rger:
> Yes, good point - I'll come up with a couple of options - maybe we can discuss it in the JF?

sure!

PR #1505, opened 2024-12-18
feat(generic): refactor GenericRDFBaseRenderer and add CIDOC renderers

b1rger:


sennierer:
What about keeping this PR open a bit longer until the upstream (PFP) serializers are a bit more stable? I am eg thinking about the fact that we dont have a simple group serializer in PFP and/or here yet.

apis_core/generic/serializers.py
```diff
@@ -87,3 +95,100 @@ def get_content_type_key(self, obj) -> str:
...
+
+        return g
+
+    def to_representation(self, instance):
+        g = Graph()
+        content_type = ContentType.objects.get_for_model(instance)
+
+        CRM = Namespace("http://www.cidoc-crm.org/cidoc-crm/")
+        g.namespace_manager.bind("crm", CRM, replace=True)
+        g.namespace_manager.bind("owl", OWL, replace=True)
```
sennierer, line 157: I think all the NS declarations, bindings and creation of the graph could be done in a dedicated function in the base class (maybe including the import of the custom namespaces such as CRM from a central location as suggested in another comment). That function could be called in init and store NS and g in an attribute. Like that CRM wouldnt be created in several places and all the inheriting serializers could already use it.

sennierer (commented):
Thanks for this!
Two more things that I thought about:
- we should create (in another PR) serializer functions for CRM to be reused. Eg a function for adding a time-span and a function to add a appellation. Both of these things are repeating everywhere (person, places, institutions etc)
- and we should add the birth and death events to the E21_Person serializer, given that there is a birth and a death date in the object.

sennierer (approved):
perfect, thanks!

PR #1530, opened 2025-01-15
feat(templates): update to bootstrap 5

b1rger:
Closes: #1529


koeaw (commented):
Having looked at what the UI looks like in TB compared to before, this is what I've noticed so far:

1. spacing between individual form fields is kind of gone, they look very squished together,
2. dropdowns don't seem to have any styling anymore,
3. with the new font, text content in table rows not being 100% vertically centred seems a bit more obvious/slightly more irritating to the eye,
4. file [`override_bootstrap.min.css`](https://github.com/acdh-oeaw/apis-core-rdf/blob/6e9b3061bc027081299ff9f7c4a554963e785ae3/apis_core/core/static/css/override_bootstrap.min.css) can/should probably be removed (as the comment indicates it was meant as temp. fix for Bootstrap 4/an issue with select2 pre Bootstrap 5)

The fact links are now underlined by default is probably a good thing (accessibility).

I've not yet investigated what the reasons for 1. and 2. are, i.e. it might not have anything to do with prior customisations on our side.

ETA: None of these are blockers for merging this PR, just wanted to make note of them.

koeaw:
Form fields before/after (dropdown included):

b1rger:
>     1. spacing between individual form fields is kind of gone, they look very squished together,
> 
>     2. dropdowns don't seem to have any styling anymore,

please install `crispy-bootstrap5` and try with 
```
CRISPY_ALLOWED_TEMPLATE_PACKS = "bootstrap5"
CRISPY_TEMPLATE_PACK = "bootstrap5"
```

koeaw:
> please install `crispy-bootstrap5` and try with

Will do!

koeaw:
Hm, I'm wondering: wouldn't it make sense to put this in the settings repo?

koeaw (approved):

b1rger:
> Should this be left to individual ontologies to fix if they don't like it?

If its a bootstrap default, yes. If its something we introduced in core, it probably makes sense to fix it in core

PR #1567, opened 2025-01-29
feat(generic): add a collections filter

b1rger:
Closes: #1557

### DIFF
--- a/apis_core/generic/filtersets.py
+++ b/apis_core/generic/filtersets.py
@@ -1,6 +1,46 @@
+import logging
+
+import django_filters
+from django.apps import apps
+from django.contrib.contenttypes.models import ContentType
 from django_filters.filterset import FilterSet
 
+from apis_core.generic.forms.fields import IncludeExcludeField
+
 from .forms import GenericFilterSetForm
+
+logger = logging.getLogger(__name__)
+
+
+class CollectionsFilter(django_filters.filters.ModelMultipleChoiceFilter):
+    """
+    A simple filter for connections to collections. It provides an `include`/`exclude`
+    selector, which allows to define what the filter should do.
+    """
+
+    @property
+    def field(self):
+        return IncludeExcludeField(super().field, required=self.extra["required"])
+
+    def filter(self, queryset, value):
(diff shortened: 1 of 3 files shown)


b1rger:
> UI observation - excuse me if it is an artifact of improperly configured bootstrap.
> The label for "collections" looks different from other form labels. In my browser it seems to take a `legend` tag(because it is enclosed within a `fieldset`?) while other form labels use the `label` tag, even though all of them use the same style class `form-label`

Ah, good catch! I had to do a lot of fiddling with the input group setup, so I totally overlooked the label. I'll have a look, maybe I can fix that

b1rger:
> UI observation - excuse me if it is an artifact of improperly configured bootstrap.
> The label for "collections" looks different from other form labels. In my browser it seems to take a `legend` tag(because it is enclosed within a `fieldset`?) while other form labels use the `label` tag, even though all of them use the same style class `form-label`

Oke, so apparently this is how crispy bootstrap5 renders fieldsets. I've now set the widgets `use_fieldset` attribute to `False`, but I'm not sure if thats the right approach - it might make sense to group those elements in a fieldset, but on the other hand its only two elements. We could also set the `label_class` attribute on the form helper, but that would have an effect on *all* labels and legends.
(Maybe there is even a way to make the `include`/`exclude` select element somehow show that it refers to the collection multiselect...)

b1rger:
> UI observation - excuse me if it is an artifact of improperly configured bootstrap.
> The label for "collections" looks different from other form labels. In my browser it seems to take a `legend` tag(because it is enclosed within a `fieldset`?) while other form labels use the `label` tag, even though all of them use the same style class `form-label`

Oke, so apparently this is how crispy bootstrap5 renders fieldsets. I've now set the widgets `use_fieldset` attribute to `False`, but I'm not sure if thats the right approach - it might make sense to group those elements in a fieldset, but on the other hand its only two elements. We could also set the `label_class` attribute on the form helper, but that would have an effect on *all* labels and legends.
(Maybe there is even a way to make the `include`/`exclude` select element somehow show that it refers to the collection multiselect...)

PR #1587, opened 2025-02-07
feat(generic): introduce generic menu entry

b1rger:
This commit introduces a template tag in the generic module that lists
all content types that inherit from the Generic Model but are not
Collections, Entities, Relations or History models.
This templatetag is then used to create a menu entry to manage those
types.

Closes: #1582

### DIFF
--- a/apis_core/generic/abc.py
+++ b/apis_core/generic/abc.py
@@ -89,6 +89,10 @@ class GenericModel:
     def get_delete_permission(self):
         return permission_fullname("delete", self)
 
+    @classmethod
+    def get_view_permission(self):
+        return permission_fullname("view", self)
+
     def get_merge_charfield_value(self, other: CharField, field: CharField):
         res = getattr(self, field.name)
         if not field.choices:
--- a/apis_core/generic/templates/base.html
+++ b/apis_core/generic/templates/base.html
@@ -1,0 +1,29 @@
+{% extends "base.html" %}
+{% load generic %}
+
+{% block main-menu %}
+  {{ block.super }}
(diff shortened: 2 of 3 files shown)


b1rger:
> @b1rger very useful to have this menu, thanks! Similar to History models, can we exclude `RootObject` from this list?

I've refactored the PR and the list now only includes content types the requesting user has **read** access to. Also, the menu entry is only shown if there is at least one content type the user as read access to.

> I would be more comfortable if the existence of RootObject is hidden from the user. For example **_Are you sure you want to delete RootObject (2)?_** is an innocuous question that the user might confidently say Yes to and then be shocked that an entity object has also quietly disappeared.

The delete button in the list view should only be visible for users that have access to the functionality. So maybe you can remove the RootObject delete permission for those users? I just tested this on a local discworld and a user does not need the RootObject permissions to create a Person.

b1rger:
> 1. Is it possible to not have the Other menu as the very first menu item? How about
>        _Entities,  Other objects, Relations, Collections, ..._ ?

yes, by changing the order of the `INSTALLED_APPS`

>  2. And could we call this menu something like "Other objects" or "Other models"?

done!

b1rger:
> Not a showstopper for this PR - but just wondering if we should make it easier to choose the order of items in the top bar.

I think this is a consequence of the template inheritance. `apis_entities` builds upon the generic template and creates `abstractentity_{detail,edit}.html` template. Those are then extended by the relations templates. It would probably make things easier if it was similar to the top bar, that all apps simply append to one template instead of having an inheritance chain. But I have currently no idea how to implement that in HTML

Issue #1397, opened 2024-11-15
background object creation failing in relation form


b1rger:
>   So I'll create a separate issue for the background object creation failing and close this one. Thanks for the answers!

_Originally posted by @gythaogg in https://github.com/acdh-oeaw/apis-core-rdf/issues/1364#issuecomment-2470491829_

This refers to the autocomplete not giving appropriate feedback if the creation of an object did not succeed

b1rger:
> Thank you, this is correct - my new issue tab is still in draft as I was investigating the slightly different ways in which many to many fields record creation in forms failed and wasn't sure if I should report different issues or a common one.

no problem, its a complicated matter as I just found out myself ;)

> This refers to the autocomplete not giving appropriate feedback if the creation of an object did not succeed

so actually this is only about the relations form autocomplete. **If** the other autocompletes have a `Create ...` option and the create fails, it shows an error message, but that error message is not really helpful. I've created #1398 to make this error message bit more meaningful.
For the relations autocomplete, this error message (which is returned by the autocomplete endpoint in the `error` field of the JSON response) somehow does not show up. I guess thats (again) a problem with our custom `reinit_select2` event listener

b1rger:
> I guess thats (again) a problem with our custom reinit_select2 event listener

we probably have to copy the `success` function:
https://github.com/yourlabs/django-autocomplete-light/blob/45a9ff6e95e58348b90b3c1f01f01cb6ca10b4d3/src/dal_select2/static/autocomplete_light/select2.js#L128

b1rger:
> @b1rger This is now fixed, right?

Well, when the creation process fails, this still does not show the error message, so I'd keep that open until we have fixed that

b1rger:
fixed now with c2b621fa620dfcbd43858e8a264f0a93a0859a08

PR #1598, opened 2025-02-13
birger/refactor autocomplete

b1rger:
- **feat(core): introduce a custom select2 function**
- **feat(generic): use customized autocomplete widgets instead of upstreams**
- **feat(relations): use customized autocomplete widgets instead of upstreams**
- **chore(deps): drop apis-override-select2js dependency**
- **feat(relations): replace custom select2 call**


b1rger:
This should fix #1397 - given that this does not use a custom select2 for the relation forms, it now should show the error message if the entity could not be created

b1rger:
> The message is displayed twice the first time I click on Create and every subsequent click on Create adds two more error messages as well.

Ah, yes, apparently the form got sent two times on every click. I've now added `e.stopImmediatePropagation();` in the event handler, fingers crossed this does not break anything.

> But would it be possible to clear previous errors and only show the latest one?

Yes, previous errors are now removed and a timestamp is added to the error, so you still see that its a "new" error, even if its the same error message.

PR #1635, opened 2025-02-26
fix(generic): remove many2many relation tables from columns

### LINKED ISSUE #1500: Additional columns - Models used as M2M fields
Potentially related to #1297. In the listview - the additional columns menu contains models that use this model as an M2M field

Example: https://discworld.acdh-ch-dev.oeaw.ac.at/sample_project.profession/

Additional column lists "Person" which appears as another ID column when added, but does not display any data.

image

gythaogg:
closes #1500

* [`apis_core/generic/views.py`](diffhunk://#diff-84062e304a3382520894171842713a262215b783bbf8ecd5488132b428511826R12): Imported `ManyToManyRel` from `django.db.models.fields.related`.
* [`apis_core/generic/views.py`](diffhunk://#diff-84062e304a3382520894171842713a262215b783bbf8ecd5488132b428511826R189): Modified the `_get_columns_choices` method to exclude fields that are instances of `ManyToManyRel`.

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -9,6 +9,7 @@ from django.contrib.auth.mixins import PermissionRequiredMixin
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.validators import URLValidator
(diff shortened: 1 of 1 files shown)


b1rger:
nice!

PR #1654, opened 2025-03-05
feat(relations): major rewrite of the relations list

b1rger:
This commit implements a major rewrite of the way the relation
views, form and templates are handled.
The original approach was very tightly coupled to the relations being
always listed in separate sections were every section was listing
relations that point to a specific content type. This approach is now
abstracted and we have now have a partial template
`relations/partials/list_relations.html` which simply lists relations
specified by a list of natural keys. The previously hardcoded "create
relation" links are now in a separate partial template.
The approach of passing the list of relevant relations as a list of
natural keys is then also used in the `ListRelations` view, which now
lists relations that refer to a specific entity and the **type** of the
relations that should be listed is passed as a list of natural keys.
The list of relevant relation natural keys is also passed from the list
to the form, which after successful submit, again passes is to the
listview (so htmx can reinclude the updated list).

This now allows us to include customized lists of relation lists, which
are simply specified by passing a list of relations natural keys. The
table that is used for those relations is not found using the target of
the relation, but can simply be passed using the `table_suffix`
parameter (but in the default template it defaults to the target).


sennierer (approved):
works nicely in my local dev environments

PR #1656, opened 2025-03-06
fix: remove duplicate entries in API

### LINKED ISSUE #1655: [api] api is delivering double entries
https://oebl-pfp.acdh-ch-dev.oeaw.ac.at/apis/api/apis_ontology.institution/?limit=200&offset=200&format=json
and https://oebl-pfp.acdh-ch-dev.oeaw.ac.at/apis/api/apis_ontology.institution/?limit=200&format=json
both contain an institution with the ID 136545

What is especially strange is that there seems to be no pattern in the duplicate entries (entries before and after are not in both) and the API overall delivers the correct number of objects. Instead of the duplicate entries others are missing that can be found with querying for the name.

sennierer:
not sure this is the right way to do it, but this fixed the issue
in the OEBL instance for institutions.

resolves #1655

### DIFF
--- a/apis_core/apis_entities/models.py
+++ b/apis_core/apis_entities/models.py
@@ -1,8 +1,10 @@
 import functools
+import logging
 import re
 
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.db.models.base import ModelBase
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.urls import NoReverseMatch, reverse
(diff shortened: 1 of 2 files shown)


sennierer:
really strange issue and this is not a clean solution (as distinct needs to be called on querysets to achieve normal behavior), but at least its called high up in the chain and should work with most of our installations.

b1rger:
> really strange issue and this is not a clean solution (as distinct needs to be called on querysets to achieve normal behavior), but at least its called high up in the chain and should work with most of our installations.

I actually think its a very clean solution! Can we maybe also add this to the listview queryset?

b1rger (changes requested):
Can we maybe also add this to the listview queryset?

b1rger (approved):

Issue #1657, opened 2025-03-06
Entity menu items sort order


koeaw:
In `v0.39.0`, the sort order of the model classes (verbose names) in the entity menu has reverted to... I imagine the order they appear in `models.py` in. Before that, they were sorted alphabetically.

Screenshot of current appearance:

<img src="https://github.com/user-attachments/assets/3143ec31-807b-467d-a3e1-c89dd0bca64e" width="200">

PR #1658, opened 2025-03-06
fix(apis_entities): reorder entity menu items

koeaw:
Changes the current sort order of entity model
classes in the entities menu in `base.html` from
(presumably) by `ContentType` `id` to alphabetical by `ContentType` `name`, which is based on the
model's `verbose_name`.

Resolves: #1657

### DIFF
--- a/apis_core/apis_entities/templates/base.html
+++ b/apis_core/apis_entities/templates/base.html
@@ -16,7 +16,7 @@
     </a>
     <div class="dropdown-menu" aria-labelledby="navbarDropdown">
       {% entities_content_types as content_types %}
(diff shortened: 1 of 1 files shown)


koeaw:
Since we display `verbose_name_plurals` in the UI, sorting by `verbose_name` (via [`ContentType` `name`](https://github.com/django/django/blob/bad1a18ff28a671f2fdfd447bdf8f43602f882c2/django/contrib/contenttypes/models.py#L148-L153)) might not always match expectations 100%, but should be an OK approximation.

koeaw:
| Before  | After |
| --- | --- |
|  |  |

PR #1669, opened 2025-03-12
birger/provide uri shortname

b1rger:
- feat(apis_metainfo): provide a shortname for Uris
- feat(apis_entities): use uri shortname in linked_open_data template


b1rger:
> Looks nice!
> 
> Instead of defaulting to "unknown" could we default to the domain name - like bdrc for bdrc.io, etc?
> 

I thought about something similar. Its not so straightforward to extract the domain name from the string, but I've now replace "unkown" with the netloc, thats longer than the domain, but a lot simpler to extract

Issue #1668, opened 2025-03-12
[rdf-imports] filter definition in toml files works only with prefixes


sennierer:
currently due to this logic: https://github.com/acdh-oeaw/apis-core-rdf/blob/681309447b37774ac76fea059518854816940423/apis_core/utils/rdf.py#L26-L32

the predicates and subjects in a filter definition need to have prefixes (`expand_curie` only works with prefixes). We need to support pure URLs as well. Suggestion is to - similar to turtle syntax - allow in toml files to define URIs in `<>` (eg `<http://test.at>`)

PR #1671, opened 2025-03-13
fix(utils): allow to pass an URI string to rdf config filter

b1rger:
Closes: #1668


sennierer (approved):
working as expected in my OEBL instance

Issue #1676, opened 2025-03-14
Relation forms do not show external autocomplete results


b1rger:
The relation refactoring done in 29d39dab2644980f51c7d09a6a24695e4b8cdde5 changed how the form gets initializes. Now we don't know anymore what the target of the relation will be, when the form gets created. Therefore the form shows a generic autocomplete instead of the one specific to the target, but the generic autocomplete does not allow for external results.

PR #1677, opened 2025-03-14
fix(relations): repair the relation target autocomplete

b1rger:
With the refactoring done in 29d39dab2644980f51c7d09a6a24695e4b8cdde5 we
don't know anymore which target type the relation create form refers to.
This leads to the creation form using a different autocomplete endpoint,
one where we don't have external autocomplete results.
This commit fixes this partially for Relations that allow only **one**
specific type of relation target.

Closes: #1676

### DIFF
--- a/apis_core/relations/forms/__init__.py
+++ b/apis_core/relations/forms/__init__.py
@@ -84,6 +84,23 @@ class RelationForm(GenericModelForm):
         """
 
         self.params = kwargs.pop("params", {})
+        # workaround: if there is only one possible subj or obj type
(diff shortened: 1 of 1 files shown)


sennierer:
Using this PR the autocomplete works again, but creating a relation using a selected entity from a autocomplete (only tried one that does not exist in the db yet) fails with `apis_core.relations.models.Relation.subj_content_type.RelatedObjectDoesNotExist: Relation has no subj_content_type.`

sennierer (approved):
working nicely. Thanks!

Issue #1678, opened 2025-03-14
Trying to access objects via /entity/ uri results in error


koeaw:
When I try to access individual objects via their (local to the project) `/entity/` uri, I'm getting the following error:

```
AssertionError at /entity/12345/
Cannot apply DjangoModelPermissions on a view that does not set `.queryset` or have a `.get_queryset()` method.
(...)
Raised during: apis_core.apis_entities.api_views.GetEntityGeneric
```

I'm assuming this is due to `queryset` having got dropped from `GetEntityGeneric` when it was rewritten in f825739.

This is what the view looks like now (same for `v0.39.0`, which I'm still on):
https://github.com/acdh-oeaw/apis-core-rdf/blob/681309447b37774ac76fea059518854816940423/apis_core/apis_entities/api_views.py#L13-L19

When I add the `queryset` back, everything works as expected (i.e. like I'm used to from Frischmuth, which is stuck on `v0.27.0`).

b1rger:
> (local to the project) `/entity/`

can you point us to the urls.py? I'd like to understand how the `APIView` is included

koeaw:
> > (local to the project) `/entity/`
> 
> can you point us to the urls.py? I'd like to understand how the `APIView` is included

TB [doesn't currently](https://github.com/acdh-oeaw/apis-instance-tbf/tree/c35411417aeeb0b9ba9634b63abb32ed0591ba8f/apis_ontology) have its own urls.py file.

I [removed](https://github.com/acdh-oeaw/apis-instance-tbf/commit/556e21341ed1f9738873ebb9d8c209ee862ba6bf) it so apis-acdhch-default-settings' urls would be used, which are pointed to by the settings repo's [`ROOT_URLCONF`](https://github.com/acdh-oeaw/apis-acdhch-default-settings/blob/9314402fd552c517c5646b244fa1bbc584128414/apis_acdhch_default_settings/settings.py#L136).

b1rger:
> > > (local to the project) `/entity/`
> > 
> > can you point us to the urls.py? I'd like to understand how the `APIView` is included
> 
> TB [doesn't currently](https://github.com/acdh-oeaw/apis-instance-tbf/tree/c35411417aeeb0b9ba9634b63abb32ed0591ba8f/apis_ontology) have its own urls.py file.
> 

Ah, sorry, then i misunderstood the "local to the project" part.


[...]

PR #1679, opened 2025-03-14
fix(apis_entities): readd queryset to view

koeaw:
`queryset` was dropped from `GetEntityGeneric` when it was rewritten in f825739, which results in an
AssertionError when trying to access objects via
their /entity/ uris.
This adds the `queryset` back (like it was before).

Closes: #1678


koeaw:
Was fixed in #1690 -> closing.

PR #1690, opened 2025-03-24
fix(apis_entities): add permission class to GetEntityGeneric view

b1rger:
Without the permission class, DRF falls back to the default permission
class, which might sometimes be the DjangoModelPermissions class. This
class, however, uses the queryset to assess the permission of the user.
Given that this view does not use a queryset and it only forwards a
request, it makes sense to simply set the permission class to allow any
request.

Closes: #1678


koeaw (approved):
Works for me!

Issue #1696, opened 2025-03-24
Can't create new relations between objects – empty dropdowns


koeaw:
When trying to create new relations for existing objects, the dropdowns for the "other" entity remain empty.

<img src="https://github.com/user-attachments/assets/6d322bb7-0133-407b-8800-e7722691ec0b" width="500">

My browser console shows the following error:

>  an Uncaught ReferenceError: yl is not defined

pointing to:

https://github.com/acdh-oeaw/apis-core-rdf/blob/d67a7826e9d95ec85533de234b4b7b6b5884774b/apis_core/relations/static/js/relation_dialog.js#L4

I'm experiencing this both on `v0.39.0`, which is the current deployed version for TB, and while testing `v0.40.1` locally before upgrading.

PR #1697, opened 2025-03-24
fix(core): include the main autocomplete_light js in the header

b1rger:
Include the `autocomplete_light.js` script in the page header.

Usually this is automatically done during form rendering, because the
file is included in the media assets of forms using autocomplete light.
And the main setup method of autocomplete light is then triggered by the
`load` event. But this event is not fired when we include dialogs via
htmx. Therefore it does not help much to include the script in the
relation forms media assets and we have to include it in the pages
header as a global javascript file.

Closes: #1696


koeaw (approved):
This fixes things for me, thx!

PR #1707, opened 2025-03-26
feat(generic): print exception message if module not found

b1rger:
This is useful for the automatic module lookup, because it tells you
that it could not import a module and might even tell you why it could
not import a module.


b1rger:
> could the debug message also show which module could not be imported

done!

PR #1711, opened 2025-03-27
fix: add lang-tags to short importer def

sennierer:


apis_core/utils/rdf.py
```diff
@@ -53,6 +53,38 @@ def find_matching_config(graph: Graph) -> dict | None:
...
 
+def build_sparql_query(filter: str) -> str:
+    """
+    Build a SPARQL query with language preferences.
+
+    Args:
+        filter: Filter sring as defined in the toml.
+                needs to include the predicate and optionally
+                a lang tag to filter for separated with a comma.
+                Eg "wdt:P122,en". Only one lang_tag per filter is supported.
```
b1rger, line 64: I would argue that the users didn't even know that at some point of this PR it was possible to have multiple language tags and therefore we can expect the user to only use one language tag. Which makes the code even simpler, because we can say 
```
lang_tag = None
if ',' in filter:
   predicate, lang_tag = filter.split(",", 1)
```
and then simply test for value of lang tag and use it as it is, without having to access the first element

apis_core/utils/rdf.py
```diff
@@ -53,6 +53,38 @@ def find_matching_config(graph: Graph) -> dict | None:
     return None
 
 
+def build_sparql_query(filter: str) -> str:
```
b1rger, line 56: `filter` is a reserved word in Python so this might lead to problems with some IDEs

apis_core/utils/rdf.py
```diff
@@ -61,9 +93,8 @@ def get_value_graph(graph: Graph, curies: str | list[str]) -> list:
         if curie.startswith("SELECT "):
             results = graph.query(curie)
         else:
-            results = graph.query(
-                "SELECT ?object WHERE { ?subject " + curie + " ?object }"
-            )
+            query = build_sparql_query(curie)
+            results = graph.query(query)
```
b1rger, line 97: we could also put the `if curie.startswith("SELECT ")` block into the `build_sparql_query` function and then simply use `results = graph.query(build_sparql_query(curie))`

apis_core/utils/rdf.py
```diff
@@ -53,6 +53,38 @@ def find_matching_config(graph: Graph) -> dict | None:
...
+            WHERE {{ 
+                ?subject {predicate} ?object 
+            }}
+        """
+    if lang_tag:
+        if len(lang_tag) > 1:
+            logger.debug(
+                f"You habe more than one lang tag in {filter}, but only one per filter definition is allowed."
+            )
+        lang_tag = f'FILTER(LANG(?object) = "{lang_tag[0]}") }}'
```
b1rger, line 81: how about using [langmatches](https://en.wikibooks.org/wiki/SPARQL/Expressions_and_Functions#LANGMATCHES)?

apis_core/utils/rdf.py
b1rger: on the other hand, I just realized that we can have simpler code **and** multiple language tags. we can simply query for multiple languages and **not** do a coalesce, because the outer logic **expects** multiple values to be returned, so for the outer logic it makes not difference if there is `["wdt:PXX,de", "wdt:PXX,en"]` or `["wdt:PXX,de,en"]`

b1rger (changes requested):

apis_core/utils/rdf.py
b1rger: Because in the end, the importer only takes the first in the list:
https://github.com/acdh-oeaw/apis-core-rdf/blob/7c5f5b7698826f213e44491d1adabbb694b0e1a4/apis_core/generic/importers.py#L111

apis_core/utils/rdf.py
```diff
@@ -53,17 +53,42 @@ def find_matching_config(graph: Graph) -> dict | None:
...
+    """
+    if curie.lower().split()[0].strip() in ["select", "prefix"]:
+        return curie
+    predicate, lang_tag = curie.split(",", 1)
+    query = f"""
+            SELECT ?object 
+            WHERE {{ 
+                ?subject {predicate} ?object 
+            }}
+        """
```
b1rger, line 77:
```
lang_tag=None
if ',' in curie:
    predicate, lang_tag = curie.split(",", 1)
```

b1rger (changes requested):

b1rger (approved):

apis_core/utils/rdf.py
```diff
@@ -53,17 +53,44 @@ def find_matching_config(graph: Graph) -> dict | None:
...
+        curie, lang_tag = curie.split(",", 1)
+    query = f"""
+            SELECT ?object 
+            WHERE {{ 
+                ?subject {curie} ?object 
+            }}
+        """
+    if lang_tag:
+        lang_tag = f'FILTER LANGMATCHES(LANG(?object), "{lang_tag}") }}'
+        query = query.replace("}", lang_tag)
```
b1rger, line 82: If we set `lang_tag = ""` we can simply add `{lang_tag}` to the query f-string instead of `if` + `replace`

sennierer: not sure about that actually. Wouldnt a filter for a empty lang tag filter for strings without any lang tag? While when no lang tag is set we want just any string, no matter if with or without lang tag.

b1rger: I meant the one assignment that is `lang_tag = None` currently. So basically:
```
lang_tag = ""
if "," in curie:
    curie, lang_tag = curie.split(",", 1)
    lang_tag = f'FILTER LANGMATCHES(LANG(?object), "{lang_tag}")'
query = f"""
            SELECT ?object 
            WHERE {{ 
                ?subject {curie} ?object {lang_tag}
            }}
        """
```

sennierer: ah, yes, that makes sense

b1rger (approved):

PR #1717, opened 2025-03-28
fix: tunnel models through importer chain

### LINKED ISSUE #1716: target class set in relations gets lost in the chain of functions when importing RDF
`GenericModelImporter`-> `create_instance()` -> relations -> `create_object_from_uri(uri, model)` -> `get_importer_for_model(model)` -> `GenericModelImporter(uri, model)` -> `request()` -> `get_something_from_uri(uri)`

`get_something_from_uri(uri)` uses only the uri to find the config while in the `GenericModelImporter` we have the target class. we should extend the `get_something_from_uri` function to accept an optional list of possible target classes. Given that we set the allowed toml definitions now in the classes this would allow to limit down the possible toml definitions. This again would not only improve speed, but also mitigate the problem to actually write useful filters for some resources as described in #1715 .

sennierer:
resolves #1716

### DIFF
--- a/apis_core/generic/importers.py
+++ b/apis_core/generic/importers.py
@@ -45,7 +45,10 @@ class GenericModelImporter:
     def request(self, uri):
         # we first try to use the RDF parser
         try:
-            data = get_something_from_uri(uri)
+            data = get_something_from_uri(
+                uri,
+                [self.model],
+            )
             return data
         except Exception as e:
             logger.debug(e)
--- a/apis_core/utils/rdf.py
+++ b/apis_core/utils/rdf.py
@@ -29,10 +29,9 @@ def resolve(obj, graph):
     return obj
 
 
-def find_matching_config(graph: Graph) -> dict | None:
-    models_with_config = [
-        model for model in apps.get_models() if hasattr(model, "rdf_configs")
-    ]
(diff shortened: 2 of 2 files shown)


apis_core/utils/rdf.py
```diff
@@ -29,10 +29,12 @@ def resolve(obj, graph):
     return obj
 
 
-def find_matching_config(graph: Graph) -> dict | None:
+def find_matching_config(graph: Graph, models: list | None = None) -> dict | None:
     models_with_config = [
         model for model in apps.get_models() if hasattr(model, "rdf_configs")
     ]
+    if models is not None:
```
b1rger, line 36:
```suggestion
```

apis_core/generic/importers.py
```diff
@@ -45,7 +45,12 @@ def clean_uri(self, uri):
     def request(self, uri):
         # we first try to use the RDF parser
         try:
-            data = get_something_from_uri(uri)
+            data = get_something_from_uri(
+                uri,
+                [
+                    self.model,
+                ],
```
b1rger, line 52: `[self.model]` would be easier t read and still adheres to ruff

b1rger (approved):

PR #1743, opened 2025-04-02
docs: add Sphinx autosectionlabel extension

koeaw:
Add Sphinx extension which converts references to
headings within the documentation to clickable links.

PR #1746, opened 2025-04-02
docs: wip

### LINKED ISSUE #1329: Docs: Link to History plugin from Ontology page
I think it would be helpful to make readers of the documentation aware of the [History plugin](https://acdh-oeaw.github.io/apis-core-rdf/features.html#history-plugin) right from the [Create your ontology](https://acdh-oeaw.github.io/apis-core-rdf/ontology.html#) page.

A one-sentence mention (+ link) at the end of the [Entities](https://acdh-oeaw.github.io/apis-core-rdf/ontology.html#entities) section would be sufficient. The idea is to let people know early, as not everybody will read through everything before they start developing.

koeaw:
Closes #1329


koeaw:
The formatting used here depends on config added in #1743.

koeaw:
> tested after rebasing with main, works!

Thanks!

PR #1759, opened 2025-04-04
birger/dlint lint

b1rger:
- **style(djlint): tell djlint to ignore H006 and H031**
- **style(templates): format templates according to djlint linter**
- **ci(linters): run djlint --link in addition to --check**


koeaw:
I haven't looked into `uv` yet, but will try to do so this week so I can review this PR properly (i.e. incl. running the commands that are run by the workflow).

Issue #1715, opened 2025-03-28
in RDF importer allow for empty filters


sennierer:
currently its not possible to have empty filters, but since we assign configs to models now and given the fact that its hard to find useful categories eg in wikidata to filter on (without the need for more than one ttl file) we should allow for empty filters. also to make it easier to find filtering possibilities we should make it possible to filter on a predicate only.
current workaround in oebl is to filter on some common fields to be present.

PR #1779, opened 2025-04-10
birger/1715 filters

b1rger:
- **feat(utils): make the `rdf.find_matching_config` method more verbose**
- **feat(utils): let config lookup return config if no filter is defined**

Closes: #1715


sennierer (approved):
works like intended. thanks!

PR #1789, opened 2025-04-14
docs(generic): update importers docstrings, comments

koeaw:
Fix grammar, spelling, punctuation; add paragraphs to docstrings, make first line a (useful) summary
line following PEP 257 recommendation.

### DIFF
--- a/apis_core/generic/importers.py
+++ b/apis_core/generic/importers.py
@@ -17,14 +17,14 @@ logger = logging.getLogger(__name__)
 
 class GenericModelImporter:
     """
-    A generic importer class
-    It provides the standard methods for importing data from
-    an URI and creating a model instance of it.
-    By default it fetches a resource, first tries to parse it using
-    our rdf parser, if that fails tries to parse it using json and
+    A generic importer class which provides methods for
+    importing data from a URI and creating a model instance from it.
+
+    By default, it fetches a resource and first tries to parse it using
+    our RDF parser. If that fails, it tries to parse it using JSON and
     then extracts the fields whose keys match the model field names.
     Projects can inherit from this class and override the default
-    methods or simple write their own from scratch.
+    methods or simply write their own from scratch.
     """
 
     model = None
@@ -52,7 +52,7 @@ class GenericModelImporter:
             return data
         except Exception as e:
             logger.debug(e)
-        # if everything else fails, try parsing json
+        # if everything else fails, try parsing JSON
         # if even that does not help, return an empty dict
         try:
             return json.loads(urllib.request.urlopen(uri).read())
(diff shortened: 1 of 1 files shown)


apis_core/generic/importers.py
```diff
@@ -17,14 +17,14 @@
 
 class GenericModelImporter:
     """
-    A generic importer class
-    It provides the standard methods for importing data from
-    an URI and creating a model instance of it.
-    By default it fetches a resource, first tries to parse it using
-    our rdf parser, if that fails tries to parse it using json and
+    A generic importer class which provides standard methods for
+    importing data from a URI and creating a model instance off it.
```
b1rger, line 21:
```suggestion
    importing data from a URI and creating a model instance of it.
```

apis_core/generic/importers.py
```diff
@@ -17,14 +17,14 @@
 
 class GenericModelImporter:
     """
-    A generic importer class
-    It provides the standard methods for importing data from
-    an URI and creating a model instance of it.
-    By default it fetches a resource, first tries to parse it using
-    our rdf parser, if that fails tries to parse it using json and
+    A generic importer class which provides standard methods for
```
b1rger, line 20:
```suggestion
    A generic importer class which provides the standard methods for
```

b1rger (changes requested):

apis_core/generic/importers.py
koeaw: "of" would be applicable if a model instance were a different _state_ or a true representation of the data. That's not the case though, is it? What we're doing is creating model instances _based on_ (some of) the data.

apis_core/generic/importers.py
koeaw: These methods aren't an existing, universally known/accepted or expected thing, though, but something we've thought up/come up with ourselves, right? The definite article only makes sense in reference to known quantities.

Same for "standard" actually.

apis_core/generic/importers.py
koeaw: "off of" or "from" are alternatives which could be substituted

apis_core/generic/importers.py
koeaw: I'm closing this as resolved based on your thumbs-up, @gythaogg..

apis_core/generic/importers.py
koeaw: > > I prefer "from"
> 
> Ok, will change it to that.

But I'm closing this as resolved/invalid now already, because "of" isn't grammatically correct either way and "from" instead of "off" is a preference.

koeaw:
@gythaogg I changed "off" to "from" and also removed the word "standard" as per https://github.com/acdh-oeaw/apis-core-rdf/pull/1789#discussion_r2041933451.

b1rger (approved):

Issue #1771, opened 2025-04-08
Additional German translations


koeaw:
Based on what I'm seeing on the deployed version of TB (I think there's no way to test translations on Discworld atm), I'd suggest to add the following additional German translations:

Imprint -> Impressum ... in footer*
User -> Benutzer:in ... in header (in front of logged-in user name)

*The imprint itself should also be provided in German, though I'm not sure there's a German version on https://shared.acdh.oeaw.ac.at/ (from where the imprint is served AFAIK).

b1rger:
> Imprint -> Impressum ... in footer*

The imprint is part of the apis-acdhch-default-settings app, so please create an issue there

koeaw:
> The imprint is part of the apis-acdhch-default-settings app, so please create an issue there

Will do!

koeaw:
One more:

The button for list filters is labelled "Submit" also in German. Suggested translation: either "Bestätigen" (for a generic, globally applicable label for submit buttons) or "Filter anwenden" (for this specific UI element).

koeaw:
> Git repository -> Git-Repository ... hover text for Git button in footer

The `title` text for the repo link has changed, see reviewed/accepted PR #1800.

New EN/DE version:
Code repository -> Code-Repository

PR #1800, opened 2025-04-15
feat(core): replace icon for code repository

koeaw:
Replace Git icon used for link to source code
of project in site footer with Material Icon
for "code".

Resolves: #1331


apis_core/core/templates/partials/footer-left.html
```diff
@@ -17,8 +17,6 @@
 {% git_repository_url as repository_url %}
 {% if repository_url %}
   <a href="{{ repository_url }}" title="Git repository">
```
sennierer, line 19:
```suggestion
  <a href="{{ repository_url }}" title="Code repository">
```
Given that we exchange the Git icon for a more general Code icon I suggest we also change the text accordingly.

sennierer (changes requested):
please change the text as suggested.

apis_core/core/templates/partials/footer-left.html
koeaw: I'm thinking just "source code" might make even more sense :thinking:, but I updated it to "Code repository" for now.

sennierer (approved):

b1rger:
Please translate the changed string

koeaw:
> Please translate the changed string

Can we please do this in a separate PR? This is meant to close a specific issue from October.

If the string already existed and only needed changing, sure it would make sense to include. But the string was only mentioned last week, in issue #1771 that's entirely about translations.

b1rger:
> If the string already existed and only needed changing, sure it would make sense to include. But the string was only mentioned last week, in issue #1771 that's entirely about _new_ translations.

Ok, if its to much to ask I'll just do it myself when I merge the PR

Issue #1754, opened 2025-04-03
Styling of "Relation" filter looks off


koeaw:
The styling of the "Relation" section in the list view filters (bottom of sidebar) is inconsistent with the rest of the filters.

Specifically:
* the field label uses a larger font size,
* the second dropdown (object autocomplete dropdown) is smaller in size (height) and has a darker border colour than the others.

Screenshot (source: [Person list view in Discworld](https://discworld.acdh-ch-dev.oeaw.ac.at/sample_project.person/)):

<img src="https://github.com/user-attachments/assets/cadd2d4f-a637-4b65-bc7c-b4b27add88b1" width="500">

sennierer:
dropdown autocomplete styled in dal css

PR #1804, opened 2025-04-16
birger/1754 select2 styling

b1rger:
- feat(core): add select2-bootstrap-5-theme to default stylesheet
- fix(core): use bootstrap5 theme for Select2 widgets

Closes: #1754


b1rger:
> The issue reports two bugs

thats a problem of the issue. separate issues should be reported as separate issues.

Issue #1326, opened 2024-10-24
Indentation for CSS


koeaw:
Re: #1324 – can we set the default indentation for CSS to 2 spaces instead of 4?

4 used to be the standard for HTML, CSS and JS "back in the day", but IME, 2 spaces is much more prevalent these days for all 3 of those. See also e.g. code examples on major CSS resources websites.

The main advantage, IMO, is having to do less horizontal scrolling, which is especially relevant once blocks/a hierarchy come into play.

I also think it would make sense to have consistency between HTML, CSS and JS, i.e. to not use 2 spaces for some of those and 4 for others.

koeaw:
Relatedly: I'm currently reviewing/adding various dev tool configs for TBF and saw that djLint has a [`format_css`](https://www.djlint.com/docs/configuration/#format-css) setting based on css-beautify (as well as `format_js` for js-beautify). Are there limitations to that which make a separate workflow the preferable option?

koeaw:
Ok, I don't remember this, I thought the idea was to change the indentation in any case.

I'll close this then.

PR #1811, opened 2025-04-16
Update CSS, JS indentation (config)

koeaw:
This PR updates:

1) the indentation applied to CSS and JS files by our `css-beautify` and `js-beautify` workflows, respectively, from the built-in default of 4 spaces to 2 spaces by setting `indent_size=2` in `.editorconfig` and making the workflows pick up the formatting rules from there and
2) the indent of existing .css and .js file from 4 to 2 spaces to conform to the updated config.

Resolves: #1326

Note: `.editorconfig` is added as replacement for setting rules individually via command line arguments in workflows because it makes the rules applied/the formatting required more transparent, centralises the management of these rules and should make it easy to integrate them into local development workflows since it's [supported by many popular IDEs](https://editorconfig.org/#pre-installed).

Unlike djLint, which we use on Django templates, [js-beautify](https://github.com/beautifier/js-beautify) does not support configuration via `pyproject.toml`, otherwise it would obviously have made sense to put the rules there.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: 2025 K Kollmann
+# SPDX-License-Identifier: MIT
+
+# formatting rules for CSS and JS files applied by css-beautify and
+# js-beautify workflows, respectively
+[*.{css,js}]
+end_of_line=lf
+indent_style=space
+indent_size=2
+insert_final_newline=true
+trim_trailing_whitespace=true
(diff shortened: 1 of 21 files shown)


b1rger:
we will not introduce another config file just for the sake of having indentation your way

sennierer:
I agree with @b1rger. The main languages of APIS are Python and HTML. We try to remove as much JS as possible, so lets please stick with the default formatting of JS and not add another config.

sennierer:
As a side-note, Python formatters (at least ruff and black) do not support editorconfig (see eg https://github.com/astral-sh/ruff/issues/1530) and given that (some?) IDEs seem to pass on editorconfig options to the cli command for formatters, adding editorconfig (that explicitly allows to configure python formatting options) could lead to inconsistent formatting across development environments.

PR #1831, opened 2025-04-23
fix(relations): add {ObjectType}RelationsTable before RelationsTable suffix

### LINKED ISSUE #1830: Relations table in detail view - matching table class order
Area: Detail view page of an entity, relations table on the main panel

When picking the matching class for table, it seems like `<SubjectType>RelationsTable` is higher up on the list than `<SubjectType><ObjectType>RelationsTable`

So when I define both these tables, it's always` <SubjectType>RelationsTable` that gets picked up, and `<SubjectType><ObjectType>RelationsTable` is never used.

Example: When I have `PersonRelationsTable` or `TibScholEntityMixinRelationsTable` (which I would use under the ALL tab) then the more specific `PersonWorkRelationsTable` (that I expect to use for the tab showing Person-Work relations) is never picked up.

**Potential fix**: I think instead of https://github.com/acdh-oeaw/apis-core-rdf/blob/fbc31ad537ad4f77b9c036d2fefe7e81615fb306/apis_core/relations/templatetags/relations.py#L63 if we have
```
        suffixes.insert(0, f"{suffix}RelationsTable")
```

then the SubjectObjectRelations table will have a higher precedence over SubjectRelationsTable

@b1rger I am not sure if this will have any other repercussions....

gythaogg:
so that the relations list table class specific to a target type has higher precendence over the relations list table class.

fixes #1830

### DIFF
--- a/apis_core/relations/templatetags/relations.py
+++ b/apis_core/relations/templatetags/relations.py
@@ -60,7 +60,7 @@ def relations_from(from_obj, relation_type: ContentType = None):
 def relations_list_table(context, relations, suffix=None):
     suffixes = ["RelationsTable"]
     if suffix:
-        suffixes.append(f"{suffix}RelationsTable")
+        suffixes.insert(0, f"{suffix}RelationsTable")
     table_modules = ()
     for suffix in suffixes:
         table_modules += module_paths(


b1rger (changes requested):
Could you change the commit message to s/ObjectType/suffix? The objecttype is used as a suffix in cores default templates, but it can be any string

b1rger:
and ideally use `...` in the beginning of the first line of the commit message body to make clear that it is a continuation of the commit message title

b1rger (approved):
(I meant the `...` in the first line of the body, not in the title, I fixed that)

Issue #1769, opened 2025-04-08
"Title Placeholder" string in title block


koeaw:
The base template's `title` block defaults to the string "Title Placeholder" when there is no `object`:

https://github.com/acdh-oeaw/apis-core-rdf/blob/dd57862509f54262454cce2eb4aa6a987e15ff2c/apis_core/core/templates/base.html#L9-L15

Projects which don't override `base.html` end up with this placeholder as page title in all list views and the default/index view, which isn't very useful -> a project would have to override the template or block _just_ to get rid of the placeholder. 

A variable for page title would help avoid having to touch the template. In lieu of that, an empty `<title>` tag seems preferable over displaying this random string. This could be achieved by removing the `else` part of the `if` statement.

b1rger:
> A variable for page title would help avoid having to touch the template.

We invested a lot of work to get rid of APIS specific variables (which we would have to document) and allow the projects to simply override templates (which is documented by Django).
But we can simply drop the content of the `title` block

koeaw:
> But we can simply drop the content of the title block

I wouldn't drop the entire content of the block because having info on individual objects displayed actually serves a purpose, unlike the placeholder. I.e. I'd only remove the `else` part, as suggested above.

b1rger:
The `Title Placeholder` string is also used for monitoring, so we wait for https://github.com/acdh-oeaw/apis-acdhch-default-settings/pull/194 to reach all the instances before we remove this

PR #1848, opened 2025-04-30
birger/1769 title

b1rger:
- fix(core): drop the fallback for the title block
- feat(generic): add default titles for detail and list templates
- feat(apis_entities): add default title for abstractentity template

Closes: #1769


koeaw (approved):
This works for me.

Tried this as-is and with a custom (local) base.html file extending the Core base template and overriding the `title` block to add a custom title whenever not in a list view or looking at a single object. Was shown the expected title(s) in the various views I checked.

PR #1913, opened 2025-06-06
birger/1012 faceting

b1rger:
- **feat(generic): introduce the InjectFacetsLimitOffsetPagination class**
- **feat(relations): introduce `get_relation_targets_from_model` function**
- **feat(apis_entities): add a `get_facets` method to AbstractEntity**


sennierer (changes requested):
Finally had a look at this and adapted the approach a bit in [ms/faceting-from-rel-side](https://github.com/acdh-oeaw/apis-core-rdf/tree/ms/faceting-from-rel-side) Its using counts and annotations in the querysets. In my tests with ÖBL data it took 1.5 seconds to count facets for 35.000 persons with much more relations. Its slightly slower than your original (roughly 1 sec) approach (even though has a bit less code) so I suggest we stick with your approach. 
However this fix here https://github.com/acdh-oeaw/apis-core-rdf/blob/ms/faceting-from-rel-side/apis_core/apis_entities/models.py#L137-L142 should be still implemented. In my understanding of the query relations connecting objects with the same content type in object and subject position would be counted twice otherwise.

Issue #1970, opened 2025-07-24
Collections filter in All Relations list view


gythaogg:
Collection filter in All relations list view would typically return no results as the objects of the Relation base class would not be added to collections. We however show the list of objects in this page as the the objects of the sublcasses so when there are no results despite relations being part of collections, it might puzzle the users.

**Suggestion 1:** 
To avoid the misleading usecase we could remove the collections filter from this page.

**Suggestion 2:** 
Search and return any objects that inherit from the Relation base class that are part of the selected collections. 

I recommend going with Suggestion 1 (removing the collection filter) and sticking to the **Collection linked objects** view for listing all entities and relations and anything else that are in a collection.

PR #1973, opened 2025-07-30
fix(relations): remove collections filter from relations filterset

b1rger:
Closes: #1970


b1rger:
> Well since I'm not able to test this pr, I should be allowed to complain here?

Oh, you're allowed to complain everywhere! ;) but you should actually be able to test this PR

> Since the link will be added to the website via this pr, I'd say that this pr shouldn't be merged until the link works.

ehm, this PR does not add a link?

PR #1988, opened 2025-08-01
fix: run flattening on m2m fields only

### LINKED ISSUE #1987: [history] Json field creates error in history diff view
unpacking of values here https://github.com/acdh-oeaw/apis-core-rdf/blob/main/apis_core/history/models.py#L71 can return more than 2 from jsonfields and thus throw an error

sennierer:
running the flattening function for rendering history diffs on jsonfields creates an exception. this PR runs the function on m2m fields only.

resolves #1987

### DIFF
--- a/apis_core/history/models.py
+++ b/apis_core/history/models.py
@@ -65,13 +65,11 @@ class APISHistoryTableBase(models.Model, GenericModel):
         containing a mapping between the model instance string representation
         and the string representations of the connected model instances.
         """
-        if isinstance(obj, list) and all(isinstance(el, dict) for el in obj):
-            ret = []
-            for el in obj:
-                key, val = el.values()
-                ret.append(str(val))
-            return ret
-        return obj
+        ret = []
+        for el in obj:
+            key, val = el.values()
+            ret.append(str(val))
+        return ret
 
     def get_diff(self, other_version=None):
         if self.history_type == "-":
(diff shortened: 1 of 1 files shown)


apis_core/history/models.py
```diff
@@ -93,11 +97,19 @@ def get_diff(self, other_version=None):
...
                 change,
-                old=self._flatten_modelchange_many_to_many(change.old),
-                new=self._flatten_modelchange_many_to_many(change.new),
+                old=change.old
+                if change.field in concrete_fields
+                else self._flatten_modelchange_many_to_many(change.old),
+                new=change.new
+                if change.field in concrete_fields
+                else self._flatten_modelchange_many_to_many(change.new),
             )
```
b1rger, line 113: I think it would be way more readable if we simply run the `.replace` method **only** on unconcrete fields

apis_core/history/models.py
```diff
@@ -65,7 +65,11 @@ def _flatten_modelchange_many_to_many(self, obj: Any) -> Any:
         containing a mapping between the model instance string representation
         and the string representations of the connected model instances.
         """
-        if isinstance(obj, list) and all(isinstance(el, dict) for el in obj):
+        if (
+            isinstance(obj, list)
+            and all(isinstance(el, dict) for el in obj)
+            and all(len(el.values()) == 2 for el in obj)
+        ):
             ret = []
```
b1rger, line 73: How does this change relate to the problem at hand?

b1rger (changes requested):
please add `history` as a scope to the commit message

apis_core/history/models.py
sennierer: simplified the approach and run the flattening on m2m fields only

apis_core/history/models.py
sennierer: removed that as it is not needed when running it on m2m fields only

b1rger (approved):

PR #1997, opened 2025-08-05
feat(metainfo): move Uri to new `metainfo` app

b1rger:


sennierer:
tested it with a local oebl instance. migrations took a while, but uris are there, new entities can be imported and manually created. looks fine to me!

PR #2041, opened 2025-08-21
feat(entities): introduce entities app

b1rger:


sennierer (commented):
With the ÖBL db this is not working.
Thats what I did:
- took a new db dump from production
- loaded the db in my local db
- did a stash and uv sync to ensure that code is on par with the server
- -> everything runs fine
- added a symlink to my local apis_core (makes debugging easier)
- checked out the branch
- added `apis_core.entities` in installed apps before `apis_core.apis_entities`
- -> app still runs fine
- run `python manage.py migrate`
- following error: 
```
ProgrammingError at /apis/apis_ontology.person/update/140928

column apis_ontology_person.rootobject_ptr_id does not exist
LINE 1: ...person" INNER JOIN "apis_metainfo_rootobject" ON ("apis_onto...
```
In the table the rootobject_ptr_id is changed to entity_ptr_id, but I am not sure where this join comes from.

sennierer (commented):
Obviously inheritance in ontology needs to be moved from AbstractEntity to Entity. After doing that app worked correct again (I think we should put the things that need to  be done in the PR description, and/or print them to stdout in the migrations)
Apart from that templates and templatetags need to be moved from apis_core.apis_entities to the new app to get the UI components back (eg Entities list in the top menu, or relations in the entity edit form). I think this should also go in this PR.

Issue #1539, opened 2025-01-21
add a list of filters used to the top of the result set (or highlight filters used)


sennierer:
> to difficult to implement, but:
> - add a list of filters used to the top of the result set (or highlight filters used) 
> - save columns used in filters in session for cookie (so it doesnt get lost on reload)
> 
> ![Image](https://github.com/user-attachments/assets/99cff36d-5945-4e1d-8bbd-26e6095178c2) 

 _Originally posted by @sennierer in [#1391](https://github.com/acdh-oeaw/apis-core-rdf/issues/1391#issuecomment-2604557240)_

PR #2046, opened 2025-08-25
birger/1539 filter values

b1rger:
- **feat(generic): embed filterset labels in a div with a css class**
- **feat(generic): add a css class for the selected filter values**

Closes: #1539


b1rger:
> 2. `#DC143C` is a bit too red, perhaps we can choose on a friendlier highlight colour that doesn't scream error?

Can you propose an alternative?

Issue #1445, opened 2024-11-25
relations: configuring pagination


gythaogg:
We can configure the pagination size of list views by setting `paginate_by` in for the table for  list views. 

In the detail view of an entity, (new) relations are paginated and we can modify the page size here by
1. adding a `per_page` parameter in the URL 
or
2. adding a `per_page` attribute to the `Meta` class of the `EntityRelationsTable`

It would be nice to have a uniform configuration method - either `per_page` in `Meta` class or `paginate_by` as an attribute of the table class for tables everywhere in APIS

b1rger:
I'd suggest to drop the `paginate_by` attribute. The `Meta.per_page` attribute is part of django-tables2. I did not know that this Meta attribute exists when I introduced the `paginate_by` setting. I think it makes more sense to use the upstream solution than a custom one.

PR #2072, opened 2025-09-05
fix(generic): drop the `paginate_by` override of tables

b1rger:
The table class already has a `Meta.per_page` attribute, so the
`.paginate_by` attribute is not needed.

Closes: #1445

BREAKING-CHANGE: if you use `.paginate_by` in your tables, please switch
to `Meta.per_page`


b1rger:
@gythaogg I'm not sure if that really fixes the issue #1445 because I'm not sure if I understood the problem correctly

PR #2090, opened 2025-09-16
birger/importer refactor

b1rger:
Refactor the import logic.
Given that we realized that the configuration for importing, the URI where we import from and the model are rather tightly coupled, it doesn't make sense to have some kind of overengineered auto lookup finding the correct importer.
This PR moves the basic import logic into the `GenericModel` and the concrete import logic for specific types into the respective abstract models.

### DIFF
--- a/apis_core/uris/signals.py
+++ b/apis_core/uris/signals.py
@@ -5,7 +5,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
-from apis_core.generic.signals import post_merge_with
+from apis_core.generic.signals import post_merge_with, pre_import_from
 from apis_core.uris.models import Uri
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


apis_core/generic/abc.py
```diff
@@ -127,6 +130,67 @@ def get_verbose_name_plural(cls):
...
+        """
+        errors = {}
+        if data:
+            for field in self._meta.fields:
+                if data.get(field.name, False):
+                    value = str(data[field.name][0])
+                    try:
+                        field.clean(self, value)
+                    except Exception as e:
+                        errors[field.name] = str(e)
```
sennierer, line 185: not sure where these errors end, but given the broad catch we should (in case the errors are only forwarded to forms) also log the exception (for when cli commands imort)

sennierer (changes requested):
Works in my test as intended, some minor suggestions regarding code duplication

apis_core/generic/abc.py
b1rger: Added a `logger.info` statement

apis_core/apis_entities/abc.py
```diff
@@ -42,12 +47,14 @@ def __str__(self):
             return f"{self.forename} {self.surname}"
         return self.forename or self.surname or force_str(_("No name"))
 
-    @classmethod
-    def rdf_configs(cls):
-        return {
-            "https://d-nb.info/*|/.*.rdf": "E21_PersonFromDNB.toml",
-            "http://www.wikidata.org/*|/.*.rdf": "E21_PersonFromWikidata.toml",
-        }
+    import_definitions = {
```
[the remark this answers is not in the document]
b1rger: done

sennierer (approved):
Thanks!

Issue #1540, opened 2025-01-21
look into replacing css `float` with `flex`


sennierer:
see how much of the page needs to be refactored to change from `float` to `flex`

b1rger:
If flex is the way to go, it would be great to have this implemented as an example for how to use flex. I'm currently waiting for this issue and postponing any bootstrap/CSS modifications, because I'm not sure if the way I implement thing might already be outdated

koeaw:
> If flex is the way to go, it would be great to have this implemented as an example for how to use flex. I'm currently waiting for this issue and postponing any bootstrap/CSS modifications, because I'm not sure if the way I implement thing might already be outdated

I'll try to look at this this week!

b1rger:
Any progress?

koeaw:
> Any progress?

I haven't forgotten about this, but sorry, not yet! 😭

koeaw:
> I have no idea how this is now actionable, what exactly should be implemented or not, so I'm adding the needs-attention label again

The above was just meant as explanation, I've already worked on this this week, just not yet submitted a PR.

PR #2114, opened 2025-09-22
feat(apis_entities): move from CSS floats to flex

koeaw:
Convert header layout of single object view from
relying on CSS `float`s to being `flex`-based.
This change makes the header section responsive
and allows for greater flexibility in terms of how elements can be arranged or displayed in individual apps, including at different screen sizes.
The new layout has Bootstrap Flex classes at its core but also uses vanilla CSS for tweaks to text, margins or visibility of elements to make it easier to override these downstream (i.e. without having to always resort to `important` rules to trump Bootstrap's `important` defaults).
Some elements are now left off small screens by default for better use of screen real estate.
Elements are primarily referenced/manipulated via class selectors rather than ID selectors to keep the calculated weight of CSS rules (specificity) low; additional "empty"/ unused IDs could always be added to elements later on to allow devs to "skip" weight calculations (by basing overrides directly on IDs, which carry the most weight).

Solves: #1540

### DIFF
--- a/apis_core/apis_entities/static/css/apis_entities.css
+++ b/apis_core/apis_entities/static/css/apis_entities.css
@@ -2,12 +2,3 @@
 #single-object-header h2 {
     text-align: center;
 }
-
-/* navigation between objects (in list) from single object view header */
-.prev-in-list {
-    float: right;
-}
-
-.next-in-list {
-    float: left;
-}
--- a/apis_core/apis_entities/templates/apis_core/apis_entities/abstractentity.html
+++ b/apis_core/apis_entities/templates/apis_core/apis_entities/abstractentity.html
@@ -16,9 +16,10 @@
       <div class="card-header">
 
         {% block card-header-content %}
-          <div id="single-object-header" class="row">
-            <div class="col-md-2">{% include "apis_entities/partials/prev_url.html" %}</div>
-            <div class="col-md-8">
+          <div id="single-object-header"
+               class="d-flex flex-column flex-md-row justify-content-between justify-content-sm-around">
+            <div class="d-flex justify-content-end">{% include "apis_entities/partials/prev_url.html" %}</div>
+            <div>
               <h2>
                 {% include "apis_entities/partials/listview_url.html" %}
                 {{ object }}
@@ -31,7 +32,7 @@
 
               </h2>
             </div>
-            <div class="col-md-2">{% include "apis_entities/partials/next_url.html" %}</div>
+            <div class="d-flex justify-content-start">{% include "apis_entities/partials/next_url.html" %}</div>
           </div>
         {% endblock card-header-content %}
 


koeaw:
The link back to the list view is now in the same place as the headline _of_ the list view, btw.

b1rger:
please split this up into several commits. the goal was to look into replacing css float with flex. everything else (changing of text-decoration of elements, changing font size of elements, ... ) seem to be additional changes? Why are they part of that POC? That clearly makes it harder to review and discuss the individual changes. Which of the changed lines are exactly about that referenced issue? If I now want to mimic the fix to the actual issue ("replace css float with flex") in other parts of the interface, I have no idea what of this commit I should build upon

koeaw:
The problem with layout changes like these is that they are complex by nature. They affect more than just a couple of attributes or elements. Switching from `float` to `flex` is not as simple as switching out a couple of attributes or elements, it's more like switching to a different API for something. 

They way HTML is structured also quickly makes things hard to review either way (because whitespace changes/changed indentation makes whole blocks appear "changed" when only a single element is added/removed).

There are also n different way in which CSS rules can be constructed. Many of them would be equivalent in any given moment, some may be chosen based on perceived _potential future advantages_... which wouldn't (also shouldn't) make it into any sort of documentation. (Meaning CSS offers limited/often random learning opportunities.)

My primary goal with layout changes is to not break the layout. Which extends to visuals/appearance. Which, in a case like this, means to make the transition from one layout type to another as seamless or smooth as possible.

To simultaneously optimise for code readability would require to either add in temporary workarounds/fixes only to immediately remove them again, or to split certain changes off into "preparatory" commits that don't do anything by themselves (akin to splitting model changes and migrations into separate commits). There is very little in here that is "extra". Basically only the added (permanent) entity type below the object name (instead of only showing the type when there's no link to the list view) and the switch from `<h2>` to `<h1>` (which is a fix of incorrect use of CSS headings, which would have required an interim fix if applied separately), I think.

I think it would make sense to demo some of what I'm talking about here in person, to work out what the best solution for frontend changes of this kind is.

b1rger:
What does "(now unused) selectors" in the commit message refer to?

b1rger:
I'm afraid I don't really understand the actual changes:
You add the `flex-column` class **and** the `flex-md-row` class, which means the elements are vertically by default but horizontally for all screens with a width ≥768px - is this on purpose?
I also don't understand why there is a `justify-content-sm-around` - why would we want to use another approach to justifying the content for screens ≥576px?
And do we actually need the flex classes in the child elements? It seems it also works if we simply drop them

b1rger:
(Also I think we can just delete the `apis_entities.css` - with the move to flex there is nothing in there that is still useful)

koeaw:
> What does "(now unused) selectors" in the commit message refer to?

That `prev-in-list` and `next-in-list` remain in place in the templates but aren't used for CSS rules anymore. But "class attributes" is probably more precise. I'll reword that.

koeaw:
> (Also I think we can just delete the `apis_entities.css` - with the move to flex there is nothing in there that is still useful)

There currently still is a rule for the heading in there. Removing it will also remove the rule.

Do you want me to move that to the template?

koeaw:
> But "class attributes" is probably more precise. I'll reword that.

Updated the commit message.

b1rger:
> > (Also I think we can just delete the `apis_entities.css` - with the move to flex there is nothing in there that is still useful)
> 
> There currently still is a rule for the heading in there. Removing it will also remove the rule.

A rule that is not useful if we use flexbox with justify-content, isn't it?

b1rger:
> > A rule that is not useful if we use flexbox with justify-content, isn't it?
> 
> `justify-content-*` only works on `flex` containers.
> 
> The `div` containing the heading currently isn't. We can remove the rule if we make that one a `flex` container as well.

You lost me.

I'm talking about:
```
#single-object-header h2 {
    text-align: center;
}
```
In current `main`, if we remove this `text-align: center`, the h2 moves to the left.
In your branch, if we remove the `text-align: center`, nothing changes, because of the `justify-content-` rules.
So why would we want to keep a css rule that does not change anything?

koeaw:
> Then please flag it as a potentially breaking change with ! in the commit message.

Wouldn't the layout change itself be considered a breaking change?

Maybe we should discuss how frontend changes should be handled in terms of how we classify them. Like, how is a frontend refactor even defined or measured. I'm not sure if it can be done in code (in useful ways) but if the visual appearance of the UI isn't used as basis either, then I don't know where the line is drawn between what can or can't be added.

I think I'll close this PR for now because with a pivot away from changing the floats to flex, the original issue this PR was meant to solve isn't applicable anymore either.

koeaw:
@b1rger I can explain/show how the text alignment matters and isn't affected by the justifying of content in person.

PR #2129, opened 2025-09-24
fix(generic): fix Django Debug Toolbar error

### LINKED ISSUE #2089: Issue with Django Debug Toolbar middleware with v0.53.0+
I'm running into an error with Django Debug Toolbar, which I use for local development, trying to [upgrade to `v0.53.0`](https://github.com/acdh-oeaw/apis-instance-tbf/pull/373).

I don't know if something about the included middlewares changed with that release (as [placement of the debug middleware matters](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#add-the-middleware)), so was wondering if, alternatively, the changes to `GenericModel` could have caused it.

So far I placed the debug toolbar middleware before all other middlewares like so:
```python
    MIDDLEWARE = [
        "debug_toolbar.middleware.DebugToolbarMiddleware",
        *MIDDLEWARE,  # noqa: F401, F405
    ]
```
which now results in an `AttributeError` for `'attname'`:

```python traceback
Internal Server Error: /
Traceback (most recent call last):
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/PATH/TO/VENV/lib/python3.13/site-packages/debug_toolbar/middleware.py", line 145, in __call__
    return self._postprocess(request, response, toolbar)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/debug_toolbar/middleware.py", line 182, in _postprocess
    panel.generate_stats(request, response)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/debug_toolbar/panels/signals.py", line 91, in generate_stats
    receiver_name = getattr(receiver, "__name__", str(receiver))
                                                  ~~~^^^^^^^^^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/apis_core/generic/abc.py", line 30, in __repr__
    return super().__repr__()
           ~~~~~~~~~~~~~~~~^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/db/models/base.py", line 590, in __repr__
    return "<%s: %s>" % (self.__class__.__name__, self)
                                                  ^^^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/db/models/base.py", line 593, in __str__
    return "%s object (%s)" % (self.__class__.__name__, self.pk)
                                                        ^^^^^^^
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/db/models/base.py", line 657, in _get_pk_val
    return getattr(self, meta.pk.attname)
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/db/models/query_utils.py", line 260, in __get__
    val = self._check_parent_chain(instance)
  File "/PATH/TO/VENV/lib/python3.13/site-packages/django/db/models/query_utils.py", line 280, in _check_parent_chain
    return getattr(instance, link_field.attname)
                             ^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'attname'
```

Commenting the debug toolbar middleware line makes the problem go away.

The order of inheritance in the TB models does not make a difference.

When I printed out some info to pinpoint the problem, `APISHistoricalRecords` showed up (via `VersionMixin`, I suppose).

Changing this line:
https://github.com/acdh-oeaw/apis-core-rdf/blob/d26d6fe85637c940b56b6057eaed44964c18b459/apis_core/generic/abc.py#L30

to `return repr(super())` fixed the problem. (Evaluates to  `<super: <class 'GenericModel'>, <APISHistoricalRecords object>>`).

koeaw:
Fix `AttributeError` thrown by Django Debug Toolbar (middleware) since version `v0.53.0.` by replacing return statement of `super().__repr__()` with
`repr(super())`.

Closes: #2089

### DIFF
--- a/apis_core/generic/abc.py
+++ b/apis_core/generic/abc.py
@@ -27,7 +27,7 @@ class GenericModel(models.Model):
     def __repr__(self):
         if id := getattr(self, "id", None):
             return super().__repr__() + f" (ID: {id})"
-        return super().__repr__()
+        return repr(super())
 
     @property
     def content_type(self):


koeaw:
> This can not be merged, it breaks important behavior of APIS/Django. I don't really see how this change would make any sense. You're replacing the `__repr__` (which usually returns information about the object) with a functionality that returns information about the type of the object. So all objects of one type have the same representation. Can you explain what the idea behind this change is?

No, I can't.

My reply to being assigned the issue was:

> I'm not sure what to do here.
> 
> I can change the line that seems to fix the issue but won't be able to provide a very useful commit message for the change because I don't fully understand why it works. I also haven't been able to find info on similar issues with the toolbar nor been able to track down other potential causes.

PR #2145, opened 2025-09-29
fix(generic): only add table columns with non-field names

### LINKED ISSUE #2143: Columns listed twice in columns selector
Testing `v0.55.0`, I'm seeing fields/columns listed twice.

Screenshots from (local) TB:

Default list view for `Person` entity:
Image

Columns selector:
Image

The duplicated fields appear to be ones that are both defined in the model (class [`Person`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/ecb012ce3c7725e524adcc1a5b56974b1b272619/apis_ontology/models.py#L271) inherts from `E21_Person`) and explicitly set on the table (via [`PersonTable`'s `fields`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/ecb012ce3c7725e524adcc1a5b56974b1b272619/apis_ontology/tables.py#L93C9-L93C15) meta variable).

Difference in behaviour from other columns in the list:
To remove a twice-listed column from the table, both occurrences in the dropdown need to be deselected.
To _readd_ a twice-listed column after it's been removed, it's sufficient to select only one of the two occurrences in the dropdown.

koeaw:
When adding custom table columns to the `choices`
columns selector, only add columns whose names do
not already appear in `choices`, i.e. columns whose names do not match model field names.
This applies to columns used for extra data but also field columns whose column names do not match the
original field names (meaning columns which point to specific model fields via "accessor", but which are not given the same names on the table).

Closes: #2143

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -196,10 +196,13 @@ class List(
         ]
         # we add any annotated fields to that
         choices += [(key, key) for key in self.get_queryset().query.annotations.keys()]
-        # lets also add the custom table fields
+        # add custom table columns whose names do not match model field names;
+        # this covers columns for extra (non-field) data but also applies to
+        # columns for fields which are set on the table using different names
         choices += [
             (key.name, capfirst(str(key) or key.name or "Nameless column"))
             for key in self.get_table().columns
+            if key.name not in [f[0] for f in choices]
         ]
         # now we drop all the choices that are listed in columns_exclude
         choices = list(filter(lambda x: x[0] not in columns_exclude, choices))


b1rger:
I think the table columns should take precedence over the model attribute colums

b1rger:
> Whats the status of this PR?

ping

koeaw:
> > Whats the status of this PR?
> 
> ping

I guess I left this mostly unsure as to whether the suggested changes would require a split into two separate PRs or if everything could still go into this one as a big `fix`, just rewritten so table columns take precedence.

Will still look at this today and update the changes until next week at the latest.

koeaw:
I might be missing something, but it appears the problem this PR was trying to solve + the changes that were suggested were already implemented in d9229e7 in December anyway. -> Closing.

PR #2153, opened 2025-09-30
Relation labels

### LINKED ISSUE #2147: relations: return name, reverse_names as lazy strings
Return the relation labels, which are the relations equivalent of verbose names in entities, as lazy strings to support internationalisation

If the subclasses define the `verbose_name` as a lazy string and override the `reverse_name(cls)` to return lazy strings then this would support internationalisation already. 

1. " reverse" added in the default reverse name labels should be translated 
⚠️ What should this be? Rückwärts? Inverse?
2. Test if `name_and_reverse_name` returns the appropriate translation requested

https://github.com/acdh-oeaw/apis-core-rdf/blob/8dbd2f56c28b56138ff3b4d7552819d0db4fc0fe/apis_core/relations/models.py#L149-L166

gythaogg:
This pull request introduces internationalization (i18n) support for relation names by using Django's translation utilities. It updates both the core relation logic and the sample project models to leverage translatable strings, and adds a German translation file for these relation names.

**Internationalization of relation names:**

* Updated `Relation` class methods (`name`, `reverse_name`, and `name_and_reverse_name`) in `apis_core/relations/models.py` to return translatable strings using `gettext` and `format_lazy`, ensuring all relation names are ready for translation. [[1]](diffhunk://#diff-7eed37546a1cb5292a1bc2308561a69856e2d08a2b43f9809737339663abc786R9-R11) [[2]](diffhunk://#diff-7eed37546a1cb5292a1bc2308561a69856e2d08a2b43f9809737339663abc786L150-R169)
* Modified all sample relation model classes (e.g., `IsCousinOf`, `IsPartOf`, `IsSiblingOf`, `LivesIn`) in `sample_project/models.py` to set their `verbose_name` as a translatable string and updated their `reverse_name` methods to return translatable strings as well. [[1]](diffhunk://#diff-6f96855dd07da1d9708d10b82324702e8c08401d1df38086448432a7950c0b27R2) [[2]](diffhunk://#diff-6f96855dd07da1d9708d10b82324702e8c08401d1df38086448432a7950c0b27R31-R75)

**Localization resources:**

* Added a German translation file (`django.po`) under `sample_project/locale/de/LC_MESSAGES/` with translations for all relevant relation names.

closes #2147


b1rger (changes requested):
Please use `--add-location file` for generating the po file (or remove all the line numbers from the existing one)

PR #2155, opened 2025-10-01
feat(generic): save cookie with form data

### LINKED ISSUE #1564: Listview: store selected columns in session data or cookie

b1rger:
Closes: #1564

### DIFF
--- a/apis_core/generic/views.py
+++ b/apis_core/generic/views.py
@@ -17,6 +17,7 @@ from django.db import transaction
 from django.db.models.fields.related import ManyToManyRel
 from django.forms import modelform_factory
 from django.forms.utils import pretty_name
+from django.http import QueryDict
 from django.shortcuts import get_object_or_404, redirect
 from django.template.exceptions import TemplateDoesNotExist
 from django.template.loader import select_template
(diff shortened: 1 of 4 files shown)


apis_core/generic/views.py
```diff
@@ -126,6 +126,23 @@ class List(
     template_name_suffix = "_list"
     permission_action_required = "view"
 
+    def setup(self, *args, **kwargs):
+        super().setup(*args, **kwargs)
+        content_type = ContentType.objects.get_for_model(self.model)
+        self.cookiename = f"{content_type.app_label}.{content_type.name}_list"
```
gythaogg: This can result in illegal cookie names if  `content_type.name` contains spaces or other delimiters that are [forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#:~:text=A%20%3Ccookie%2Dname%3E%20can%20contain%20any%20US%2DASCII%20characters%20except%20for%20control%20characters%20(ASCII%20characters%200%20up%20to%2031%20and%20ASCII%20character%20127)%20or%20separator%20characters%20(space%2C%20tab%20and%20the%20characters%3A%20(%20)%20%3C%20%3E%20%40%20%2C%20%3B%20%3A%20%5C%20%22%20/%20%5B%20%5D%20%3F%20%3D%20%7B%20%7D))
For example: 
image

gythaogg (changes requested):
Works well when the list view names do not contain spaces or special characters. Thanks!

Use `content_type.model` instead of `content_type.name` to avoid illegal cookie names

gythaogg (approved):
Works!
This might drastically change the user experience though - if projects do not want this feature, it would be nice to have the possibility to disable it via a django setting. 

Other options might be to have the option to remember/not the form data as 
1. a user profile setting in the database
2. a session setting stored in cookies

Issue #2209, opened 2025-10-29
When no importer is configured for a source - after the toast error the user is left in a blank page


gythaogg:
For example - if https://bad-source.com is a LOD URI for [Rincewind](https://discworld.acdh-dev.oeaw.ac.at/sample_project.person/1), clicking on the **... merge data** on the URI would take me to https://discworld.acdh-dev.oeaw.ac.at/sample_project.person/enrich/1?uri=https://bad-source.com

Here I am shown the following error as a toast message

> Import not configured for URI https://bad-source.com

But once it disappears I am left with a blank page. 

## Expected behaviour:

Toast error message shown on the detail page after redirecting the user back to the detail page in merge data  fails.

PR #2211, opened 2025-10-29
feat(apis_entities): for merging linked URIs go via to merge_or_enrich

b1rger:
This way we can check if the URI can even be merged

Closes: #2209


b1rger:
Try together with #2202

b1rger:
> User clicks on merge data on a LOD link that does not have an import configuration - show the toast error on the detail page

That should not even be possible. I've now added a helper method that lists URIs that are importable and the `... merge with` link now only exists for those URIs.

gythaogg:
Okay, I'd argue that if the merge link is not available then it doesn't support import. 

But - that's just my opinion. I'll approve this PR as it is.

gythaogg (approved):

Issue #2218, opened 2025-10-31
Shrunk import form


gythaogg:
Import form used to look like:

<img width="1912" height="281" alt="Image" src="https://github.com/user-attachments/assets/74a87844-8944-4194-b805-395b21e4cbf6" />

but it [now](https://discworld.acdh-dev.oeaw.ac.at/sample_project.person/import) sits narrow and awkwardly on the left of the page, with blank space on the right

<img width="1912" height="281" alt="Image" src="https://github.com/user-attachments/assets/e9340e40-64eb-43c9-9f52-7ae6eadec678" />

If this is intentional, what is the rationale behind this?

b1rger:
Not intentional, its a bug

b1rger:
It's due to 6ec9fd8. The first column now only uses as much space as it needs

PR #2221, opened 2025-11-02
fix(generic): override content instead of col in import template

b1rger:
The `col` block is targeted at sidebars and uses only as much space as
the inner html needs. Given that the import form does is not part of a
sidebar, it makse sense to not use the `col` and instead overload the
whole `content` block

Closes: #2218


gythaogg (approved):
Can confirm that create/view/edit forms now take up the content block instead of the sidebar.

Issue #2219, opened 2025-10-31
Places from Wikidata without a "default for all languages" are imported without a label


gythaogg:
This can be confusing for the user as the user is shown a label in the autocomplete.

Examples: 
[Free imperial city of Ravensburg](https://www.wikidata.org/wiki/Q2423057)
[Imperial City of Rothenburg](https://www.wikidata.org/wiki/Q2423057)

PR #2234, opened 2025-11-25
Allow to pass data from autocomplete to importers

b1rger:
- **feat(generic): allow to pass additional data to import URIs**
- **feat(utils): allow ExternalAutocompleteAdapter to append data to URI**
- **feat(utils): append query data parameter in autocomplete classes**

Closes: #2219

### DIFF
--- a/sample_project/querysets.py
+++ b/sample_project/querysets.py
@@ -13,7 +13,8 @@ class GroupExternalAutocomplete(ExternalAutocomplete):
             params={
                 "filter": "type:CorporateBody",
                 "format": "json:preferredName,geographicAreaCode,dateOfEstablishment,broaderTermInstantial",
-            }
+            },
+            data_mapping={"label": "label"},
         ),
     ]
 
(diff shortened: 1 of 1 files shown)


gythaogg (commented):
The URI saved should be without query params
image

gythaogg (changes requested):
The URI is still stored with query params when I select a value from the autocomplete in the Enrich/Merge usecase
image
 
image

gythaogg (approved):
Can confirm that with this PR, when the external endpoint doesn't return information, a fall back label from autocomplete is applied to surname in Person and label in Group and Places (tested on sample_project).
Importing directly from a URL works like it used to.

Minor caveat - if a person only has a first name and no surname (example: https://d-nb.info/gnd/7505655-0) then the surname is again populated with the fallback label from autocomplete. This is an edge case though and doesn't block the PR from being merged.
Thanks @b1rger!

PR #2274, opened 2026-01-08
birger/save column cookie2

b1rger:
- **feat(generic): add a reset checkbox to column and filterset forms**
- **feat(generic): store list form data in cookie**

### DIFF
--- a/apis_core/generic/forms/__init__.py
+++ b/apis_core/generic/forms/__init__.py
@@ -21,13 +21,15 @@ logger = logging.getLogger(__name__)
 
 
 class ColumnsSelectorForm(forms.Form):
+    columns = forms.MultipleChoiceField(required=False)
+    remember = forms.BooleanField(required=False, label="Remember selected columns")
+
     def __init__(self, *args, **kwargs):
         choices = kwargs.pop("choices", [])
         super().__init__(*args, **kwargs)
         initial = kwargs.get("initial", {}).get("choices", [])
-        self.fields["columns"] = forms.MultipleChoiceField(
-            required=False, choices=choices, initial=initial
-        )
+        self.fields["columns"].choices = choices
+        self.fields["columns"].initial = initial
         self.helper = FormHelper()
         self.helper.form_method = "GET"
         self.helper.form_tag = False
--- a/apis_core/generic/templates/generic/genericmodel_list.html
+++ b/apis_core/generic/templates/generic/genericmodel_list.html
@@ -41,6 +41,14 @@
             {% crispy filter.form filter.form.helper %}
           {% endblock filter %}
 
+          <div id="div_id_filterset-remember" class="form-check">
+            <input type="checkbox"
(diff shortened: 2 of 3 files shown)


gythaogg (commented):
1. It would be nice if the Remember columns gets automatically checked whenever the user makes a change to the selected columns **always** OR **only when _Remember filters_ is selected**
2. When the user selects Remember filters once, the next time the user visits the page, the filters are loaded correctly, and Remember filters checkbox checked - so if the user changes the filters, the original filter that the user consciously "saved" is overwritten by the current filter. I think it is possible that the user may not expect this behaviour.

apis_core/generic/templates/generic/genericmodel_list.html
```diff
@@ -41,16 +41,21 @@
...
+          </div>
           <input type="submit"
                  name="submit"
                  value="Submit"
                  class="btn btn-primary"
                  id="submit-id-submit">
         </form>
       </div>
-      <div class="card-footer text-muted">
-        <a class="btn btn-outline-secondary"  href=".">{% translate "Reset filter" %}</a>
```
gythaogg, line 52: Why is this removed as part of this PR?

b1rger: I've readded the reset button

b1rger:
> 1. It would be nice if the Remember columns gets automatically checked whenever the user makes a change to the selected columns **always** OR **only when _Remember filters_ is selected**

I thought about this and I'd be more comfortable if the functionality would be rolled out as is first, so we can see if it works as intended, before we activate the remember on change.

>  2. When the user selects Remember filters once, the next time the user visits the page, the filters are loaded correctly, and Remember filters checkbox checked - so if the user changes the filters, the original filter that the user consciously "saved" is overwritten by the current filter. I think it is possible that the user may not expect this behaviour.

I don't think there is anything we can do about this?

apis_core/generic/templates/generic/genericmodel_list.html
gythaogg: Looks good to me!

apis_core/generic/forms/__init__.py
```diff
@@ -21,13 +21,15 @@
 
 
 class ColumnsSelectorForm(forms.Form):
+    columns = forms.MultipleChoiceField(required=False)
+    remember = forms.BooleanField(required=False, label="Remember columns settings")
```
gythaogg: Can we reword "Remember columns settings" to "Remember selected columns" if it is not too much trouble? Everything works and looks good to me

gythaogg (changes requested):

apis_core/generic/forms/__init__.py
b1rger: done

gythaogg (approved):
Thank you!

PR #2283, opened 2026-02-12
poc: search

b1rger:


gythaogg:
Thank you! It looks very good. I've deployed it to the demo instance: https://apis-demo.acdh-dev.oeaw.ac.at/apis/search

Here are my initial observations in the order of my discovery, not importance:

1. apis-instances should add `apis_core.search` to `INSTALLED_APPS`
2. apis-instances should add `register` decorator from `apis_core.search.registry` to the models that should be included in search. **Aside**: _maybe we should think about keeping this decorator name more specific rather than "register"  - I'm doing `from apis_core.search.registry import register as search_registry_register`_
3. Migrations didn't run successfully initially. I got:
`django.db.utils.ProgrammingError: operator class "gin_trgm_ops" does not exist for access method "gin"`
4. After running `CREATE EXTENSION pg_trgm;` in the dbshell, migrations run successfully
5. If no models are "registered" then we get an error in the search view after clicking on Submit - this should perhaps be handled with grace (a warning to the user before they search - search hasn't been properly configured, or something like it, maybe?)
6. If `initialize_search` management command is not run, then the search does not give any results
7. Birger's comment: initialize_search should also show a warning when invoked without registering any model for search
8. By default - no content types are selected, so results from all content types are returned. Perhaps we should keep "select all" set as the default value. 
9. There is a colon at the end of the entity link in the search results - but nothing following it. Might be because this is a work in progress - I am adding it here just as an observation.
10. When I search for a value from an "Other" model, I get only the other model instance in the search result, and not any of the entities that contain that specific value as its attribute. For example: [searching for "politician"](https://apis-demo.acdh-dev.oeaw.ac.at/apis/search?query_str=politician&submit=Submit) which is a Profession - mapped as a M2M field to a Person it would be nice to see all the Persons who have this in their Profession field.
11. New records are automatically added to the search index
12. Once a record is deleted, it still appears in the search results - and the links take me to 404
13. Search scenarios with edits: 
  - Create a Person "Karina Kahr"
  - Create Person lives in Place relation
  - Change the last name to "Zahr"
  - Search for Kahr
      Expected: no results
      Current behaviour: the lives in relation shows up with the original name in its string representation
  - Search for Zahr
      Expected: Both the entity and the lives in relation
      Current behaviour: Only the entity shows up, the relation still contains the old name so it doesn't turn up in the results
  - Create a new relation and search for Zahr
      Expected: the entity and all its relations 
      Current behaviour: The relations created before the renaming no longer appear.

b1rger:
>     2. apis-instances should add `register` decorator from `apis_core.search.registry` to the models that should be included in search. **Aside**: _maybe we should think about keeping this decorator name more specific rather than "register"  - I'm doing `from apis_core.search.registry import register as search_registry_register`_

The decorator is now part of a class and that class can be imported from registry:
```
from apis_core.search.registry import search

@search.register
class MyModel(): ...
```
This is similar to how the django-auditlog registry works.


>     3. Migrations didn't run successfully initially. I got:
>        `django.db.utils.ProgrammingError: operator class "gin_trgm_ops" does not exist for access method "gin"`

Thats fixed now, the migrations contain the `TrigramExtension`


>     5. If no models are "registered" then we get an error in the search view after clicking on Submit - this should perhaps be handled with grace (a warning to the user before they search - search hasn't been properly configured, or something like it, maybe?)

I wasn't able to reproduce that.

>     7. Birger's comment: initialize_search should also show a warning when invoked without registering any model for search

Thats now implemented.


>     8. By default - no content types are selected, so results from all content types are returned. Perhaps we should keep "select all" set as the default value.

Good point, I'll look into it.

>     9. There is a colon at the end of the entity link in the search results - but nothing following it. Might be because this is a work in progress - I am adding it here just as an observation.

I removed that. The plan was to highlight the content of the model instance that contains the search string. But we do fuzzy matching, so there are model instances that don't contain exactly the search string...

>     10. When I search for a value from an "Other" model, I get only the other model instance in the search result, and not any of the entities that contain that specific value as its attribute. For example: [searching for "politician"](https://apis-demo.acdh-dev.oeaw.ac.at/apis/search?query_str=politician&submit=Submit) which is a Profession - mapped as a M2M field to a Person it would be nice to see all the Persons who have this in their Profession field.

Yes, there is documentation missing. Search uses the JSON representation of the model instance by default, but you can override that from `.search_fields()` from the instance. That is where you add the profession. We still have to implement a way to update the searchentry once the m2m field changes (i.e. the profession gets renamed)

>     12. Once a record is deleted, it still appears in the search results - and the links take me to 404

There is now a signal to delete the deleted instances searchentry.

b1rger:
Another update. There is now a separate commit which adds documentation.

> Yes, there is documentation missing. Search uses the JSON representation of the model instance by default, but you can override that from `.search_fields()` from the instance. That is where you add the profession. We still have to implement a way to update the searchentry once the m2m field changes (i.e. the profession gets renamed)

This changed. The values of ManyToMany fields are now included in the serialization of a model. You can set `m2m_fields` as argument to the decorator to define if the SearchEntry should also be updated if the target of the ManyToMany fields gets updated (i.e. if the Profession gets renamed. This can be intense if 50k instance have this profession, therefore its optin)

I've also added a caption to the search result page that explains the numbers. The results are ordered following the numbers (and then the id).

gythaogg:
> There is now a signal to delete the deleted instances searchentry.

This works (for entities, I guess relations are a work in progress).

b1rger:
> > There is now a signal to delete the deleted instances searchentry.
> 
> This works (for entities, I guess relations are a work in progress).

Can you explain once more what you mean by that? What exactly does not work, what would you expect?

gythaogg:
> > > There is now a signal to delete the deleted instances searchentry.
> > 
> > 
> > This works (for entities, I guess relations are a work in progress).
> 
> Can you explain once more what you mean by that? What exactly does not work, what would you expect?

Say, there is an entity and it has a few relations - when this entity is deleted, it no longer shows up in the search results. 
However, the relations of the deleted entity still show up in the search results and I think when an entity is modified or deleted, the relations that are present in the search index with the string representation of the entity should also be modified or deleted. The edit scenario is described in greater detail in point 13 of my [earlier comment](https://github.com/acdh-oeaw/apis-core-rdf/pull/2283#issuecomment-4488207925).

b1rger:
> > > > There is now a signal to delete the deleted instances searchentry.
> > > 
> > > 
> > > This works (for entities, I guess relations are a work in progress).
> > 
> > 
> > Can you explain once more what you mean by that? What exactly does not work, what would you expect?
> 
> Say, there is an entity and it has a few relations - when this entity is deleted, it no longer shows up in the search results. However, the relations of the deleted entity still show up in the search results and I think when an entity is modified or deleted, the relations that are present in the search index with the string representation of the entity should also be modified or deleted. The edit scenario is described in greater detail in point 13 of my [earlier comment](https://github.com/acdh-oeaw/apis-core-rdf/pull/2283#issuecomment-4488207925).

Oke, thanks. I'm not sure if this is currently possible in a clean way. When a model instance is updated, we don't know which Relations point to that model. When an instance is deleted, we update the relation, but we do that using the [update()](https://github.com/acdh-oeaw/apis-core-rdf/blob/9818fe512d977e50be3fc64a0ce93a54dc3a97e9/apis_core/relations/signals.py#L55) method, so the search update is not triggered. I think we will have to find a generic way to trigger updates in "related" objects ("related" not as in `apis_core.relation` but as in Django `GenericRelation`)

gythaogg:
A cop out might be to reinitialise the search index periodically and flag that the results might not be most up to date

apis_core/search/models.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        s.search_vector = (
+            SearchVector("title", weight="A")
+            + SearchVector("title_slug", weight="B")
+            + SearchVector("content_text", weight="C")
+        )
+        s.save()
+
+    @classmethod
+    def delete(cls, instance):
```
sennierer: This should be called different to not overwright the models own delet function: https://github.com/django/django/blob/stable/6.0.x/django/db/models/base.py#L1291

apis_core/search/models.py
```diff
@@ -0,0 +1,178 @@
...
+            lookups["content_text__ilike"] = 5
+
+        q = Q()
+        w = []
+        for lookup, value in lookups.items():
+            q |= Q(**{lookup: query_str})
+            w += [When(**{lookup: query_str, "then": value})]
+            w += [When(**{lookup: query_str_slug, "then": value - 1})]
+
+        words = (query_str + query_str_slug).split()
```
sennierer:
```suggestion
        words = (query_str + " " + query_str_slug).split()
```
if `query_str` and `query_str_slug` are stripped this otherwise results in a combination of the last word from `query_str` and the first from `query_str_slug`

apis_core/search/models.py
```diff
@@ -0,0 +1,178 @@
...
+            rankmatch=Case(When(rank__gt=0.5, then=1), default=0),
+        )
+
+        lookups = {
+            "title__contains": 10,
+            "title__ilike": 9,
+            "title_slug__contains": 8,
+            "title_slug__ilike": 7,
+        }
+        if with_content:
```
sennierer, line 86: When using content search the rank is always 5, no matter if the case sensitive word appears in the text or not. Suspected the ordering of the lookups to be the reason, but might be wrong given that it works for title.

apis_core/search/models.py
```diff
@@ -0,0 +1,178 @@
...
+        the whole `query_str` as well as the individual parts and
+        the parts slugified. Those checks are then also used as an
+        overall filter, so only results containing those strings are
+        looked at and returned.
+        """
+        query_str_slug = slugify(query_str).replace("-", " ")
+
+        query = super().get_queryset()
+        query = query.annotate(
+            rank=SearchRank(F("search_vector"), SearchQuery(query_str)),
```
sennierer:
```suggestion
            rank=SearchRank(F("search_vector"), SearchQuery(query_str, search_type="websearch", config=SET_LANGUAGE),
```
we discussed this already. websearch might not be a good idea given that we combine searches and the search syntax provided by websearch is not working for the other lookups. Nonetheless we could think of allowing to configure the language. Thats obviously not important for names and more structured data, but for text (such as the ÖBL bios that makes a difference)

sennierer: looks like not setting the language in the postgres vector search uses english as a fallback.

apis_core/search/signals.py
```diff
@@ -0,0 +1,60 @@
...
+                ):
+                    for inst in reverse_manager.all():
+                        logger.debug("Updating SearchEntry for %s", repr(inst))
+                        SearchEntry.reindex_model_instance(inst)
+
+
+@receiver(m2m_changed)
+def m2m_create_or_update_search_entry(
+    sender, instance, action, reverse, model, pk_set, using, **kwargs
+):
```
sennierer, line 53:
```suggestion
):
    if action not in ("post_add", "post_remove", "post_clear"):
        return
```
`m2m_changed` signal is triggered `pre` and `post` -> do something only on `post` actions
https://docs.djangoproject.com/en/6.0/ref/signals/#django.db.models.signals.m2m_changed

apis_core/search/registry.py
```diff
@@ -0,0 +1,28 @@
+class SearchRegistry:
```
sennierer, line 1: As discussed already, it might make sense to use the `SearchRegistry` to connect signals for all models that are registered in the global search. Similar to the first example in the [signals docs](https://docs.djangoproject.com/en/6.0/topics/signals/)

apis_core/search/management/commands/initialize_search.py
```diff
@@ -0,0 +1,45 @@
...
+        if content_types:
+            content_types = [ct.split(".") for ct in content_types]
+            content_types = [
+                ContentType.objects.get_by_natural_key(*ct) for ct in content_types
+            ]
+            content_types = [ct.model_class() for ct in content_types]
+
+            not_registered = set(content_types) - set(registered_models)
+            if not_registered:
+                print(
```
sennierer: maybe use `self.stdout.write` (https://docs.djangoproject.com/en/6.0/howto/custom-management-commands/#module-django.core.management) instead of `print`

apis_core/search/urls.py
```diff
@@ -0,0 +1,5 @@
+from django.urls import path
+
+from . import views
+
```
sennierer, line 4: add `app_name = "search"` or similar to scope urls

sennierer (commented):
Working nicely with ÖBL data. Two notes:
- we might want to reconsider how the ranking is done. In the current setup when I search for "Krieg" I get a person named "Kriegelsteiner" way above people who have "Krieg" mentioned in their Bios several times
- given that the search gets slower the more tokens someone enters it could be an option to limit the `contains` based search part to a certain number of tokens?  And run the full query only through vector serach?

b1rger:
> * given that the search gets slower the more tokens someone enters it could be an option to limit the `contains` based search part to a certain number of tokens?  And run the full query only through vector serach?

The problem is that we can not **filter** based on the result of the vector search (the "rank") because that is very compute heavy. That means we have to **always** do a *dumb* filter using `contains`/`ilike` in addition to the vector search, otherwise we always get all instances as results

@sennierer @gythaogg I've refactored the search method and the indexing method, there is now an additional searchvectorfield for the content; the result page now also tries to highlight the matches. please take a look if time permits with a focus on the results + results order. You'll have to recreate the model + index. (`migrate search zero` and then `migrate search` and `initialize_search`)

apis_core/search/urls.py
```diff
@@ -0,0 +1,7 @@
...
+app_name = "search"
```
gythaogg: Since we make use of  `apis_core` like so: https://github.com/acdh-oeaw/apis-acdhch-default-settings/blob/a38fa297572af3b86f7f0ff8abd43d905079f7f7/apis_acdhch_default_settings/urls.py#L14-L15

we would have been able to access this endpoint with 
`{% url 'apis:search' %}` if there was no "search" namespace.

With the search namespace, we need to use it like
`{% url 'apis:search:search' %}` 

I prefer the former approach, without a search namespace - just putting it out here for discussion.

b1rger: I've removed the `app_name`

b1rger:
@sennierer @gythaogg I've now marked the PR as non-draft. It would be great if it could have a review before everyone goes on holidays, otherwise this whole thing will have to wait until September (which would block a lot of other work)

docs/source/administration/search.md
```diff
@@ -0,0 +1,77 @@
...
+@search.register(m2m_fields={"profession"})
+class Person(models.Model):
+    forename = models.CharField()
+    surname = models.CharField()
+    profession = models.ManyToManyField(Profession)
+```
+In this case, when a profession is changed, the `SearchEntry`s of all `Person`s
+that point to that profession are also updated.
+
+## Management commant
```
sennierer:
```suggestion
## Management command
```

apis_core/search/models.py
```diff
@@ -0,0 +1,175 @@
...
+from django.db.models.functions import Greatest
+from django.db.models.lookups import PatternLookup
+
+from apis_core.search.utils import get_search_serialization_from_instance
+
+from .utils import split_query_string_for_search
+
+
+@Field.register_lookup
+class ILike(PatternLookup):
```
sennierer, line 24: realized that this is available in all fields when.
so either we should scope it to the search module/class and/or make a short note in the doc about it (and what it does)

apis_core/search/signals.py
```diff
@@ -0,0 +1,52 @@
...
+    logger.debug("Updating SearchEntry for %s", repr(instance))
+    SearchEntry.reindex_model_instance(instance)
+
+    # Update all related models on direct m2m changes
+    for field in instance._meta.many_to_many:
+        related_manager = getattr(instance, field.name)
+        if search.is_registered(related_manager.model):
+            for inst in related_manager.all():
+                logger.debug("Updating SearchEntry for %s", repr(inst))
+                SearchEntry.reindex_model_instance(inst)
```
sennierer, line 31: not sure, but I think this is redundant with the `m2m_create_or_update_search_entry` signal receiver. If the m2m is not changed there is no need to reindex, if it is the `m2m..,,` signal will trigger and start the index. (just my reading, didnt test it)

docs/source/administration/search.md
```diff
@@ -0,0 +1,77 @@
+# Search
+
+The [apis_core.search] module provides a way to create a search index over
+various models. It uses a combination of simple search queries and PostgreSQLs
```
sennierer, line 4: I think we should make it more explicit that this only works with Postgres

apis_core/search/models.py
```diff
@@ -0,0 +1,175 @@
...
+
+    content_text = models.TextField(null=True)
+
+    title_search_vector = SearchVectorField()
+    content_search_vector = SearchVectorField()
+
+    objects = SearchManager()
+
+    class Meta:
+        unique_together = ["content_type", "object_id"]
```
sennierer: Django doc suggests to use [UniqueConstraint](https://docs.djangoproject.com/en/6.0/ref/models/constraints/#django.db.models.UniqueConstraint) instead of `unique_together`: https://docs.djangoproject.com/en/6.0/ref/models/options/#unique-together

apis_core/search/templatetags/search.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@register.filter
+@cache
+def id_to_content_type(value):
+    return ContentType.objects.get(pk=value)
+
+
+@register.simple_tag
+def get_model_field_by_content_type_id_and_name(
```
sennierer, line 20: this could be cached as well?

apis_core/search/static/css/search.css
```diff
@@ -0,0 +1,56 @@
...
+    table-layout: fixed;
+    width: 100%;
+
+    td {
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    td:first-child: {
```
sennierer:
```suggestion
    td:first-child {
```
I guess thats a typo?

sennierer (approved):
works very nicely with the ÖBL data. The UI looks much cleaner and the issues with the annotation of the search (numbers telling which filter method hit) is resolved. I have only added some minor things that I saw.

apis_core/search/models.py
b1rger: The docs contain the autogenerated reference which will list this class with its docstring. Should we make the docstring more clear?

apis_core/search/signals.py
b1rger: No, they do different things. I've added a docstring explanation. The m2m trigger only runs if the m2m field changes. The save trigger only runs if the instance is saved. Both can be triggered without the other (`.save()` does not trigger m2m changes, `person1.profession.add(profession2)` does not trigger the save signal).
The `m2m_create_or_update_search_entry` was not called at all, I've fixed that now

apis_core/search/forms.py
```diff
@@ -0,0 +1,26 @@
...
+def get_models():
+    return {
+        ContentType.objects.get_for_model(i).id: i.get_verbose_name_plural()
+        for i in search.get_registered_models()
+    }
+
+
+class SearchForm(forms.Form):
+    query_str = forms.CharField()
+    content_types = forms.MultipleChoiceField(choices=get_models(), required=False)
```
gythaogg, line 19: Can the values be selected by default - since that is the default behaviour?

gythaogg (commented):
Looks great!

The highlighter is a nice touch - but bootstrap seems to add a 3px padding around `mark` and we should get rid of it so that when a substring is highlighted there is no space around the highlighted text within the word.

The ranking is very helpful to know - but possibly too much information for our users - could we hide it in the title of ℹ️ or 🛈. If we choose to do this, then it probably makes sense to also remove the explanation for these numbers and only keep it in the documentation.

As any good search should do, this pulls up results that are partially or very slightly related to the query string, and this is reflected in the ranking displayed next to the result. I am wondering if it might be a good idea to sequester low quality matches away, behind something like _click here to see loosely related results_ - that the user can see if they REALLY want to see and not only solely on pagination to hide these low ranking results. This is relevant when a query string includes terms that occur frequently across the database and are unlikely to be relevant to the search query. 

 And just FYI - I have not tested this with custom serializers

gythaogg (commented):
Looks great!

The highlighter is a nice touch - but bootstrap seems to add a 3px padding around `mark` and we should get rid of it so that when a substring is highlighted there is no space around the highlighted text within the word.

The ranking is very helpful to know - but possibly too much information for our users - could we hide it in the title of ℹ️ or 🛈. If we choose to do this, then it probably makes sense to also remove the explanation for these numbers and only keep it in the documentation.

As any good search should do, this pulls up results that are partially or very slightly related to the query string, and this is reflected in the ranking displayed next to the result. I am wondering if it might be a good idea to sequester low quality matches away, behind something like _click here to see loosely related results_ - that the user can see if they REALLY want to see and not only solely on pagination to hide these low ranking results. This is relevant when a query string includes terms that occur frequently across the database and are unlikely to be relevant to the search query. 

Could we also include collections, uri etc as the ontologies themselves won#t be able to do that

And could you add the search tables to the admin view so it would be easier to debug/lookup? Thank you!

b1rger:
I've incorporated the requested changes. The search module now indexes any Django model that has `Config.index_for_search` set to `True`. This is the default for `GenericModels`.
Following m2m fields is now configured using `Config.search_follow_m2m`.
Changing the default fields that should be indexed is done via `Config.search_fields`.

I haven't found a way yet to select all the content types by default, but I didn't look very hard. Selecting all values would be different behavior than in other apis forms, so I think we should either change that everywhere or leave it as it is.

apis_core/search/signals.py
```diff
@@ -0,0 +1,65 @@
...
+    about the changed m2m field yet. It is also useful if the m2m field is changed
+    programmatically and the `save()` method isn't run at all.
+    """
+    if action in ("post_add", "post_remove", "post_clear"):
+        for inst in model.objects.filter(pk__in=pk_set):
+            logger.debug(
+                "Updating SearchEntry for %s (m2m_create_or_update_search_entry)",
+                repr(inst),
+            )
+        SearchEntry.reindex_model_instance(inst)
```
sennierer: I guess thats a typo? `SearchEntry.reindex_model_instance(inst)` is outside the for loop, so only the last instance gets reindexed?

docs/source/admin-guide/07-search.md
```diff
@@ -0,0 +1,97 @@
...
+    surname = models.CharField()
+    profession = models.ManyToManyField(Profession)
+
+    class Config(GenericModel.Config):
+        search_follow_m2m = {"profession"}
+```
+In this case, when a profession is changed, the `SearchEntry`s of all `Person`s
+that point to that profession are also updated.
+
+## Management commant
```
sennierer:
```suggestion
## Management command
```

docs/source/admin-guide/07-search.md
```diff
@@ -0,0 +1,97 @@
...
+For every model instance an [`SearchEntry`][apis_core.search.models.SearchEntry] will be added
+to the database, containing representations of the model instance optimized for
+search. The [`SearchSerializer`][apis_core.search.serializers.SearchSerializer] class is used by
+default to serialize the instance for search. It is based on
+[django.core.serializers.json.Serializer] but for m2m fields it adds the values
+instead of simply the keys.
+
+To control which fields of the model should be serialized for search (and thus control
+if the model instances come up in search results), you can set the `search_fields` parameter
+n the `Config` subclass:
```
sennierer:
```suggestion
in the `Config` subclass:
```

apis_core/search/management/commands/initialize_search.py
```diff
@@ -0,0 +1,57 @@
...
+            content_types = [ct.model_class() for ct in content_types]
+            models = set(models) & set(content_types)
+
+            not_configured = set(content_types) - set(models)
+            if not_configured:
+                self.stdout.write(
+                    "Models are not configured for search: "
+                    + ", ".join(map(str, not_configured))
+                )
+            models = set(models) & set(content_types)
```
sennierer: I guess either this or line 32 can go

apis_core/search/templatetags/search.py
```diff
@@ -0,0 +1,61 @@
...
+@functools.cache
+def id_to_content_type(value):
+    return ContentType.objects.get(pk=value)
+
+
+@register.simple_tag
+@functools.cache
+def get_model_field_by_content_type_id_and_name(
+    content_type_id, field_name: str
+) -> str:
```
sennierer: typing is wrong: this is returning the field itself and not the `field_name` (str)

```suggestion
) -> Field: # from django.db.models import Field
```

sennierer (changes requested):
tested it with ÖBL data and it works great and fast.
found only one thing that needs change (call to index command outside of loop in signal), the rest is typos and minor things.

PR #2341, opened 2026-04-02
fix(utils): copy values from import config

### LINKED ISSUE #2339: Import error when trying to reimport a just deleted entity object with relation
In my project, I created a class [`WorkFromGND`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/ab286d4f111a2552f805fdeece34ba9395f234c3/apis_ontology/rdf_configs/work.py#L4) in file `work.py` to import works from GND. This class is configured [to import related `Persons`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/ab286d4f111a2552f805fdeece34ba9395f234c3/apis_ontology/rdf_configs/work.py#L15) (authors of works) as well. I also added an autocomplete adapter, [`WorkExternalAutocomplete`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/ab286d4f111a2552f805fdeece34ba9395f234c3/apis_ontology/querysets.py#L31).

I've tested that imports of `Works` + author `Persons`  work this way. My go-to test case is [Die Klosterschule](https://d-nb.info/gnd/4412291-3) by [Barbara Frischmuth](https://d-nb.info/gnd/118693727).

However, I noticed that I'm getting an error when I try to do the following (_locally_, I haven't tested this on the live site so far):
1) import the above work (+ and the related author) via the `Work` autocomplete dropdown – which works fine,
2) then _delete_ both the newly created `Work` and `Person` objects – which also works*,
3) then try to import the same work _again_ (again from the autocomplete dropdown).

This results in a red error message below the import/autocomplete field which reads:
```
Could not import https://d-nb.info/gnd/4412291-3?anero_ac_data=eyJ0aXRsZSI6IFsiRGllIEtsb3N0ZXJzY2h1bGUgfCBGcmlzY2htdXRoLCBCYXJiYXJhIHwgMTk2OCJdfQ==: Expected SelectQuery, found '?' (at char 66), (line:4, col:17)
```

From step 3), the (debug) log reads as follows:

```
[...:00,503] "GET /apis/apis_ontology.work/import HTTP/1.1" 200 17287
[...:04,596] "GET /apis/apis_ontology.work/autocomplete/externalonly?q=die%20klosterschule HTTP/1.1" 200 885
[...:06,263] apis_core.generic.abc DEBUG: abc.fetch_from:165 Fetch from https://d-nb.info/gnd/4412291-3?anero_ac_data=eyJ0aXRsZSI6IFsiRGllIEtsb3N0ZXJzY2h1bGUgfCBGcmlzY2htdXRoLCBCYXJiYXJhIHwgMTk2OCJdfQ==
[...:07,039] apis_core.utils.rdf DEBUG: rdf.graph_matches_config:101 Using <class 'apis_ontology.rdf_configs.work.WorkFromGND'> for parsing graph
[...:07,039] apis_core.utils.rdf DEBUG: rdf.build_sparql_query:134 Generated SPARQL query: 
            SELECT ?object 
            WHERE { 
                ?subject owl:sameAs ?object 
            }
        
[...:07,044] apis_core.utils.rdf DEBUG: rdf.build_sparql_query:134 Generated SPARQL query: 
            SELECT ?object 
            WHERE { 
                ?subject gndo:preferredNameForTheWork ?object 
            }
        
[...:07,048] apis_core.utils.rdf DEBUG: rdf.build_sparql_query:134 Generated SPARQL query: 
            SELECT ?object 
            WHERE { 
                ?subject https://d-nb.info/gnd/118693727 ?object 
            }
        
[...:07,068] "POST /apis/apis_ontology.work/import HTTP/1.1" 200 28445
```

I didn't dig into the code which mentions `anero_ac_data`  but the one exception I found in that context starts out with "Could not _decode_..." so seems unrelated.

I can reimport the work + author fine after stopping and restarting the development server.

I'm still on version `v0.61.0` (about to upgrade) but tried this locally with `v0.62.0` too and I'm getting the same error.

Is this related to my config?
Or to do with `anero_ac_data`?

*I deleted: the relationship between the two objects + both entity objects separately + also checked there were no `Uris` left for either.

b1rger:
If we don't copy the value from the import config, we might end up
overwriting the attributes of the original config class.

Closes: #2339

### DIFF
--- a/apis_core/utils/rdf.py
+++ b/apis_core/utils/rdf.py
@@ -5,6 +5,7 @@ import inspect
 import logging
 import tomllib
 from collections import defaultdict
+from copy import copy
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, List, Tuple, Union
@@ -82,11 +83,11 @@ def graph_matches_config(graph: Graph, configfile: Path) -> dict:
             value = getattr(configfile, key)
             match value:
                 case Filter():
-                    config["filters"].append(value.value)
+                    config["filters"].append(copy(value.value))
                 case Attribute():
-                    config["attributes"][key.lower()] = value.value
+                    config["attributes"][key.lower()] = copy(value.value)
                 case Relation():
-                    config["relations"][value.name] = value.value
+                    config["relations"][value.name] = copy(value.value)
     else:
         config = load_path(configfile)
     for _filter in config.get("filters", [[(None, None)]]):


koeaw (approved):
This fixes the problem for me, thx.

PR #2345, opened 2026-04-08
fix 2344 - django-tables2 render

### LINKED ISSUE #2344: django-tables2 from v2.9 breaks tables
There's a breaking change in django-tables2 from [2.8 to 2.9](https://github.com/jieter/django-tables2/compare/v2.8.0...v2.9.0#diff-112d7cd445c59ef34d3622ea5d215a55eaa302055132f19b4ea2ebef60e70b54) . 

Their TemplateColumn.render signature changed from 
`def render(self, record, table, value, bound_column, **kwargs):`
to `def render(self, table, **kwargs):`

So in our list views (if we have the actions column) we would get `TemplateColumn.render() got multiple values for argument 'table'`

gythaogg:
chore(deps): bump django-tables2 to >=2.9 to address render() signature change
fix(tables): update render() overrides for django-tables2 ≥2.9 compatibility
fixes #2344


gythaogg:
breaking change fixed [upstream](https://github.com/jieter/django-tables2/compare/v2.9.0...v3.0.0)

PR #2356, opened 2026-04-30
birger/2088 choices

### LINKED ISSUE #2088: Relations filter in Entities search form
The relations filter to filter entities currently lists all relations, including those where the entity cannot be a part of, for example "is sibling of" in the Place search form. It would be more useful to only show those relations that the entities in this list view are allowed to be a part of.

Image

b1rger:
- **feat(relations): introduce model parameter to RelationField**
- **feat(apis_entities): limit relation types to relevant ones**

Closes: #2088

### DIFF
--- a/apis_core/relations/forms/fields.py
+++ b/apis_core/relations/forms/fields.py
@@ -5,16 +5,19 @@ from apis_core.core.fields import ApisListSelect2
 from apis_core.relations.utils import relation_content_types
 
 
-def list_relation_choices():
-    return [(rel.id, rel.model_class().name()) for rel in relation_content_types()]
+def list_relation_choices(model=None):
+    return [
+        (rel.id, rel.model_class().name())
+        for rel in relation_content_types(any_model=model)
+    ]
 
 
 class RelationMultiWidget(forms.MultiWidget):
     template_name = "relations/relation_multiwidget.html"
     use_fieldset = False
 
-    def __init__(self, attrs=None):
-        relation_choices = [(None, "---")] + list_relation_choices()
+    def __init__(self, attrs=None, choices=[]):
+        relation_choices = [(None, "---")] + choices
         widgets = [
             forms.widgets.Select(choices=relation_choices),
             ApisListSelect2(url=reverse("apis_core:apis_entities:autocomplete")),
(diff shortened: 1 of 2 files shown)


apis_core/relations/forms/fields.py
```diff
@@ -5,16 +5,19 @@
 from apis_core.relations.utils import relation_content_types
 
 
-def list_relation_choices():
-    return [(rel.id, rel.model_class().name()) for rel in relation_content_types()]
+def list_relation_choices(model=None):
+    return [
+        (rel.id, rel.model_class().name())
+        for rel in relation_content_types(obj_model=model)
```
gythaogg: In Persons filters I see 
is cousin of
is parent of
is sibling of

But I should also be able to filter Persons who are part of a particular group or live in a particular place, correct?

gythaogg (changes requested):
1. My understanding for having this filter is to be able to filter an entity based on the relations it is a part of. Example usecase: _Find all Persons born between 1100 and 1300 who lived in Bukhara_. At the moment it lists only those relations where the object model matches that of the Listview model. This means I can only filter Persons by
     - is cousin of
     - is parent of
     - is sibling of
And not by
     - lives in
     - is part of

We should be able to filter the persons by relations regardless of whether the the model is Subject or Object. 

2. In the detail page we show the reverse name of a relation treating the model of the entity as the "subject". Likewise in the relation filter we should treat the model of the Listview as the subject and show the forward or reverse name as applicable for the relation. For Discword this means:
Groups relation filter should show the relation name in the drop down  as _consists of_  and not as _is part of_ and the dropdown list should show the list of Persons
The counterpart dropdown value in Persons relation filter should show _is part of_ and show the list of Groups in the dropdown values.
Places relation filter should show _has inhabitant_ and not _lives in_. 

I hope this makes sense.

apis_core/relations/forms/fields.py
b1rger: fixed

b1rger:
> We should be able to filter the persons by relations regardless of whether the the model is Subject or Object.

Fixed

 
>   2. In the detail page we show the reverse name of a relation treating the model of the entity as the "subject". Likewise in the relation filter we should treat the model of the Listview as the subject and show the forward or reverse name as applicable for the relation. For Discword this means:
>        Groups relation filter should show the relation name in the drop down  as _consists of_  and not as _is part of_ and the dropdown list should show the list of Persons
>        The counterpart dropdown value in Persons relation filter should show _is part of_ and show the list of Groups in the dropdown values.
>        Places relation filter should show _has inhabitant_ and not _lives in_.
> 
> 
> I hope this makes sense.

It does, but it would require a major rewrite of this relation filter. I'd say that is out of scope for this specific issue

gythaogg:
> > 2. In the detail page we show the reverse name of a relation treating the model of the entity as the "subject". Likewise in the relation filter we should treat the model of the Listview as the subject and show the forward or reverse name as applicable for the relation. For Discword this means:
> >    Groups relation filter should show the relation name in the drop down  as _consists of_  and not as _is part of_ and the dropdown list should show the list of Persons
> >    The counterpart dropdown value in Persons relation filter should show _is part of_ and show the list of Groups in the dropdown values.
> >    Places relation filter should show _has inhabitant_ and not _lives in_.
> > 
> > I hope this makes sense.
> 
> It does, but it would require a major rewrite of this relation filter. I'd say that is out of scope for this specific issue

Okay, I assumed that there was already a template/util function that returned relation name in the appropriate "direction". Will create a new issue.

gythaogg (approved):

PR #2370, opened 2026-05-06
ci(check-static): look for leading slashes in static template tag

### LINKED ISSUE #2013: Add a CI check that fails on static files with leading slashes
See #2010 and #1339

b1rger:
The argument for the {% static 'foo/bar.ext' %} template tag should
not start with a leading slash because that leads to errors in some
setups.

Closes: #2013

### DIFF
--- a/.github/workflows/check-static-template-tag.yml
+++ b/.github/workflows/check-static-template-tag.yml
@@ -1,0 +1,17 @@
+#SPDX-FileCopyrightText: 2026 Birger Schacht
+#SPDX-License-Identifier: MIT
+
+name: Check for leading slashes in static template tags
+on:
+  pull_request:
+    paths:
+      - "**.html"
+
+jobs:
+  check-slash-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Look for leading slashes
+        run: |
+          ! grep -r 'static "/' apis_core


.github/workflows/check-static-template-tag.yml
```diff
@@ -0,0 +1,15 @@
+#SPDX-FileCopyrightText: 2026 Birger Schacht
+#SPDX-License-Identifier: MIT
+
+name: Check for leading slashes in static template tags
+on:
+  pull_request:
```
koeaw, line 6: If this check is only meant for templates, it could be limited to only HTML files to not have it run on every PR.

I did this in TB to speed up linting because it always took forever to run both Ruff and djLint on _everything_ when that wasn't actually necessary most of the time. See e.g. [`djlint-linter.yml`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/1e0681aad17f64fe17f7fdde2c591c5d9ce05252/.github/workflows/djlint-linter.yml#L6-L11), [`ruff-formatter.yml`](https://github.com/acdh-oeaw/apis-instance-tbf/blob/1e0681aad17f64fe17f7fdde2c591c5d9ce05252/.github/workflows/ruff-formatter.yml#L6-L11).

b1rger: done

PR #2377, opened 2026-05-18
feat(collections): add missing translations for filter labels

koeaw:
Translate labels for form fields for filtering list view by `Collections`.
Uses `pgettext` rather than `gettext` to add contextual markers to force the included translations (overrides a built-in translation for one – and only one – of the words used, which is defaulted to otherwise).

### DIFF
--- a/apis_core/collections/filters.py
+++ b/apis_core/collections/filters.py
@@ -4,6 +4,7 @@ import django_filters
 from django.apps import apps
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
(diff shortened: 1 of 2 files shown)


koeaw:
Making a note of what I said here: https://github.com/acdh-oeaw/apis-core-rdf/pull/2380#issuecomment-4486848390 re: labels using imperative voice:

> The include/exclude for Collections currently act the same way and are actually missing a colon to differentiate them visually/to make them more "grammatically sound".

though that's outside the scope of this PR.

PR #2379, opened 2026-05-18
feat(generic): be explicit about + translate filter form button

koeaw:
Adapt label for filter form "submit" button to clarify its purpose (compare: merge, delete,... buttons) and translate it to German.

### DIFF
--- a/apis_core/generic/locale/de/LC_MESSAGES/django.po
+++ b/apis_core/generic/locale/de/LC_MESSAGES/django.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-05-18 05:17-0500\n"
+"POT-Creation-Date: 2026-05-18 05:34-0500\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -139,6 +139,10 @@ msgstr "Importieren"
 #, python-format
 msgid "List %(verbose_name_plural)s"
 msgstr "%(verbose_name_plural)s Liste"
+
+#: apis_core/generic/templates/generic/genericmodel_list.html
+msgid "Apply filters"
+msgstr "Filter anwenden"
 
 #: apis_core/generic/templates/generic/genericmodel_list.html
 msgid "Reset filter"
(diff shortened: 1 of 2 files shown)


b1rger:
The form does not only send the filters but also the selected columns, so "Apply filters" might be a bit misleading

koeaw:
> The form does not only send the filters but also the selected columns, so "Apply filters" might be a bit misleading

True. I'll think about it some more.

koeaw:
I've had a look at this again and considering we've been referencing filters in the _reset_ button of the form since at least 2024, it should be fine to do the same for the submit button. The reset button also affects the column selection.

Since we don't communicate anywhere what the sidebar is for, nor properly differentiate between the section for display options vs. filter options, I'd keep the wording for the buttons.

I'm also not sure users necessarily or primarily think of the filter functionality as "filtering" – I think my users consider it more a "search" tool. So in their minds, "filter" might just be our (dev) terminology for the whole thing (i.e. everything in the sidebar, which ultimately manipulates the view, whether that's objects returned or columns displayed).

koeaw:
Ideally though, we'd separate the display options + filter form completely in the future.

koeaw:
The current labels for the Reset button (IMO the plural form makes more sense given our form(s), so the English here should be pluralised):

https://github.com/acdh-oeaw/apis-core-rdf/blob/3998b4a9c8c7401f3b801555d35bfbfc4c30774f/apis_core/generic/locale/de/LC_MESSAGES/django.po#L143-L145

b1rger:
> we've been referencing filters in the _reset_ button of the form since at least 2024

that seems to be a bug

apis_core/generic/locale/de/LC_MESSAGES/django.po
```diff
@@ -140,6 +140,10 @@ msgstr "Importieren"
 msgid "List %(verbose_name_plural)s"
 msgstr "%(verbose_name_plural)s Liste"
 
+#: apis_core/generic/templates/generic/genericmodel_list.html
+msgid "Apply filters"
```
sennierer, line 144:
```suggestion
msgid "Apply"
```

apis_core/generic/locale/de/LC_MESSAGES/django.po
```diff
@@ -140,6 +140,10 @@ msgstr "Importieren"
...
+msgstr "Filter anwenden"
```
sennierer, line 145:
```suggestion
msgstr "Anwenden"
```

apis_core/generic/templates/generic/genericmodel_list.html
```diff
@@ -53,7 +53,7 @@
           </div>
           <input type="submit"
                  name="submit"
-                 value="Submit"
+                 value="{% translate "Apply filters" %}"
```
sennierer, line 56:
```suggestion
                 value="{% translate "Apply" %}"
```

sennierer (changes requested):

PR #2380, opened 2026-05-18
feat(generic): add explicit label for columns selector

koeaw:
Add `label` for filter form select element for columns to allow for a German translation to be added.
Use wording in both English and German which makes clear(er) that the column selection applies to the list display. (Just "columns" is ambiguous since the select element is part of the filter form and could also be understood to affect how filters are applied or displayed.)

### DIFF
--- a/apis_core/generic/forms/__init__.py
+++ b/apis_core/generic/forms/__init__.py
@@ -21,7 +21,10 @@ logger = logging.getLogger(__name__)
 
 
 class ColumnsSelectorForm(forms.Form):
-    columns = forms.MultipleChoiceField(required=False)
+    columns = forms.MultipleChoiceField(
+        required=False,
+        label=_("Visible columns"),
+    )
     remember = forms.BooleanField(required=False, label="Remember selected columns")
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


b1rger (changes requested):
The string should not contain a colon (none of the other form labels do)

koeaw:
> The string should not contain a colon (none of the other form labels do)

Hmm, that's because it's an imperative rather than a descriptor of the element (contents), which the label's unusual positioning – adjacent to rather than atop the field – is suggestive of, and for which it can be kept shorter that way.  The include/exclude for Collections currently act the same way and are actually missing a colon to differentiate them visually/to make them more "grammatically sound".

Without it, I'd change it to "visible columns"/"sichtbare Spalten" and would suggest a subsequent repositioning of the label to prevent the select from getting pushed onto the next line more frequently due to the increased string length. Though the latter might make sense either way to make this label/field combo more consistent with the rest.

koeaw:
> Without it, I'd change it to "visible columns"/"sichtbare Spalten"

Updated the wording (+ the commit message, ever so slightly).

b1rger (approved):

PR #2384, opened 2026-05-20
fix(relations): make relation validation do its job

### LINKED ISSUE #2383: Bug - Validation of subj/obj content types in Relations is not working
The code for validation doesn't look quite right to me - surely we should raise an error only when _self.subj_content_type is **NOT** subj_model:_ and _if obj_model and self.obj_content_type is **NOT** obj_model:_ 
https://github.com/acdh-oeaw/apis-core-rdf/blob/bf5029aea376219a140ecc9504513bf952b94b03/apis_core/relations/models.py#L126-L133

And while fixing this, it's worth exploring if  `clean` is a better place to do the validation rather than `save` - as we know `save` will be bypassed in `bulk_create`

In Discworld, Lives In relation requires subj to be a Person and obj to be a Place in Discworld, however I was able to create all these from the backend using `LivesIn.objects.create`

Image

b1rger:
The comparison was inverted and it addition it compared two different
types (the content type and the model class), so it did not do
validation at all.

Closes: #2383

### DIFF
--- a/apis_core/relations/models.py
+++ b/apis_core/relations/models.py
@@ -125,10 +125,10 @@ class Relation(GenericModel, models.Model, metaclass=RelationModelBase):
 
     def save(self, *args, **kwargs):
         subj_model = getattr(self, "subj_model", None)
-        if subj_model and self.subj_content_type is subj_model:
+        if subj_model and self.subj_content_type.model_class() is not subj_model:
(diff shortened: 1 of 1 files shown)


gythaogg (approved):
Thanks. Works as intended.

PR #2470, opened 2026-08-13
fix(entities): tighten the wikidata place label lookups

### LINKED ISSUE #2288: Enrich Place from a Wikidata source
While import from Wikidata creates a Place with the correct label, enriching a Place entity with a wikidata source proposes to update the label with the entire object in a string representation. For example:

Image

Steps to reproduce:

1. Import https://d-nb.info/gnd/4248397-9
2. Enrich the created entity with the wikidata link  under Linked Open Data section

b1rger:
Instead of fetching *all* the labels in the rdf, we only fetch the
labels that refer to a subject that has coordinates in the file

Closes: #2288

### DIFF
--- a/apis_core/entities/rdfconfigs/place.py
+++ b/apis_core/entities/rdfconfigs/place.py
@@ -41,10 +41,32 @@ class E53_PlaceFromWikidata:
 
     label = Attribute(
         [
-            "rdfs:label,de",
-            "rdfs:label,en",
-            "wdt:P1448/rdfs:label",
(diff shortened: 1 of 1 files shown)


gythaogg (approved):
Looks good, and can confirm that it only pulls the labels that are actually labels for the place in question. Not sure what would happen when I try to enrich a Place in Wikidata that doesn't have coordinates on file - but I haven't been able to find an example for this so this is probably a moot point.

